PR #132, opened 2020-03-05
added sum method for agg data layer

grantbuster:


grantbuster:
@MRossol, deprecated py3.6. All tests pass, so cool :)

PR #133, opened 2020-03-09
Gb/output version

grantbuster:


reV/handlers/outputs.py
```diff
@@ -119,15 +119,14 @@ def writable(self):
...
         Returns
         -------
-        bool
+        writable : bool
             Flag if mode is writable
         """
+        writable = True
         mode = ['a', 'w', 'w-', 'x']
         if self._mode not in mode:
-            raise HandlerRuntimeError('mode must be writable: {}'
```
grantbuster: @MRossol, note that I removed this exception. Not sure why we raised an error here.

MRossol: No idea, probably changed the usage along the way and forgot to remove...

reV/handlers/outputs.py
```diff
@@ -111,15 +119,14 @@ def writable(self):
...
-                                      .format(mode))
-
-        return True
+            writable = False
+        return writable
```
MRossol: super nit-picky but can you add a line before the return for read-ability

grantbuster: Done and also changed the temp var to is_writable so it doesn't look conflicted with the property/method name.

PR #135, opened 2020-03-10
Resource updates/refactor

MRossol:
- change dsets property to datasets for clarity
- add __iter__
- add __contains__

### DIFF
--- a/reV/handlers/resource.py
+++ b/reV/handlers/resource.py
@@ -223,6 +223,7 @@ class Resource:
         self._meta = None
         self._time_index = None
         self._str_decode = str_decode
+        self._i = 0
 
     def __repr__(self):
         msg = "{} for {}".format(self.__class__.__name__, self.h5_file)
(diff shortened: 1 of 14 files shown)


reV/handlers/resource.py
```diff
@@ -259,6 +260,27 @@ def __getitem__(self, keys):
...
+        dset = self.datasets[self._i]
+        self._i += 1
+
+        return dset
+
+    def __contains__(self, dset):
+        test = dset in self.datasets
+        if not test:
+            msg = "{} does not exist in {}".format(dset, self)
+            raise HandlerKeyError(msg)
```
grantbuster: I think this is supposed to return False instead of raise an exception?

reV/handlers/resource.py
```diff
@@ -305,7 +327,7 @@ def h5(self):
         return h5
 
     @property
-    def dsets(self):
```
grantbuster: Thoughts on maintaining both  for backwards compatibility? I dont think keeping two aliases around would be a bad idea, especially when its a minor issue like abbreviation.

reV/handlers/resource.py
MRossol: hmm, all I changed was the property name...

reV/handlers/resource.py
MRossol: good call. I'll add dsets back in

PR #136, opened 2020-03-10
Thermal solar sam

mikebannis:
I have questions about some of the processing performed in handlers/sam_resource.py. Particularly, lines 320-327 https://github.com/NREL/reV/blob/8faf3abe29e97d0810ebfe63842b2dbbe4abc33c/reV/handlers/sam_resource.py#L320 are actively changing the input data. This may be causing some of the mismatches between reV and the SAM GUI.


grantbuster:
@mikebannis, regarding the pressure unit conversion, this should not impact the SAM execution, as you export the weather csv with the converted units and tested that in both the GUI and PySAM. SAM wants mbar/hpa so this makes sense.

reV/SAM/SAM.py
```diff
@@ -231,6 +231,8 @@ class RevPySAM:
 
     def __init__(self):
         self._pysam = self.PYSAM.new()
+        # self._pysam = self.PYSAM.default('PhysicalTroughIPHNone')
```
grantbuster: Remove this.

reV/SAM/generation.py
```diff
@@ -624,6 +625,256 @@ def default(self):
         return self._default
 
 
+class SolarThermal(Solar):
+    """ Base class for solar thermal """
+    def __init__(self, resource=None, meta=None, parameters=None,
+                 output_request=None):
+        """Initialize a SAM solar water heating object.
```
grantbuster: Incorrect docstring.

reV/SAM/generation.py
```diff
@@ -624,6 +625,256 @@ def default(self):
...
+
+        """
+        self.timezone = None
```
grantbuster: Don't use timezone as an attribute. Not necessary, only used once, and inconsistent with the other gen objects.

reV/SAM/generation.py
```diff
@@ -624,6 +625,256 @@ def default(self):
...
+        resource : pd.DataFrame
+            2D table with resource data. Available columns must have var_list.
+        """
+        self.time_interval = self.get_time_interval(resource.index.values)
+        # Create solar resource file
+        self._pysam_w_fname = self._create_pysam_wfile(self._meta, resource)
+        # pylint: disable=E1101
+        self[self._pysam_weather_tag] = self._pysam_w_fname
+
+    def _create_pysam_wfile(self, meta, resource, drop_leap=True):
```
grantbuster: Drop leap should be false by default.

reV/SAM/generation.py
```diff
@@ -624,6 +625,256 @@ def default(self):
...
+        ----------
+        meta : pd.DataFrame
+            1D table with resource meta data.
+        resource : pd.DataFrame
+            2D table with resource data. Available columns must have var_list.
+        drop_leap : bool
+            Drops February 29th from the resource data. If False, December
+            31st is dropped from leap years.
+        """
+        fname = f'{self._site}_weather.csv'
```
grantbuster: consider using .format() instead of format strings to be consistent with the rest of reV.

reV/SAM/generation.py
```diff
@@ -624,6 +625,256 @@ def default(self):
...
+        m['Elevation'] = m.elevation
+        m['Local Time Zone'] = m.timezone
+        m['Local Time Zone'] = m.timezone
+        m['Dew Point Units'] = 'c'
+        m['DHI Units'] = 'w/m2'
+        m['DNI Units'] = 'w/m2'
+        # m['GHI Units'] = 'w/m2'
+        m['Temperature Units'] = 'c'
+        m['Pressure Units'] = 'mbar'
+        # m['Wind Direction Units'] = 'Degrees'
```
grantbuster: Please clean up commented code lines!!

reV/SAM/generation.py
```diff
@@ -624,6 +625,256 @@ def default(self):
...
+        m = pd.DataFrame(meta).T
+        self.timezone = m.timezone
+        m['Source'] = 'NSRDB'
+        m['Location ID'] = meta.name
+        m['City'] = '-'
+        m['State'] = m.state.apply(lambda x: '-' if x == 'None' else x)
+        m['Country'] = m.country.apply(lambda x: '-' if x == 'None' else x)
+        m['Latitude'] = m.latitude
+        m['Longitude'] = m.longitude
+        m['Time Zone'] = m.timezone
```
grantbuster: This is why I think the self.timezone attr is unnecessary.

reV/SAM/generation.py
```diff
@@ -624,6 +625,256 @@ def default(self):
...
+                    'urban', 'population', 'landcover', 'latitude',
+                    'longitude'], axis=1)
+        m.to_csv(fname, index=False, mode='w')
+
+        # Process data
+        df = resource.copy()
+        df['dt'] = df.index
+
+        # Drop days for leap year before adjusting timezone
+        if df.dt.dt.is_leap_year.all():
```
grantbuster: This is pretty messy. Use the SAM.drop_leap() static method instead. Example from reV_run():

```
        # Use resource object iterator
        for res_df, meta in resources:

            # drop the leap day
            if drop_leap:
                res_df = cls.drop_leap(res_df)

File: reV/SAM/generation.py [hunk 9]
```diff
@@ -624,6 +625,256 @@ def default(self):
...
+        df['DNI'] = df.dni
+        df['DHI'] = df.dhi
+        df['Dew Point'] = df.dew_point
+        df['Temperature'] = df.air_temperature
+        df['Pressure'] = df.surface_pressure
+        df['Wind Speed'] = df.wind_speed
+        df = df.drop(['dt', 'dni', 'dhi', 'wind_speed', 'air_temperature',
+                      'dew_point', 'surface_pressure'], axis=1)
+        df.to_csv(fname, index=False, mode='a')
+        assert df.shape[0] == 8760 * 2
```
grantbuster: Don't do this. We run with non-30-minute solar data all the time.

reV/SAM/generation.py
```diff
@@ -624,6 +625,256 @@ def default(self):
...
+        """Calculate mean capacity factor (fractional) from SAM.
+
+        Returns
+        -------
+        output : float
+            Mean capacity factor (fractional).
+        """
+        net_power = self['annual_field_energy'] \
+            - self['annual_thermal_consumption']  # kW-hr
+        name_plate = self['q_pb_des'] * 8760 * 1000  # q_pb_des is in MW
```
grantbuster: q_pb_des is a scalar value? Not an array? *8760 is just to go from MW to MWh right?

reV/SAM/generation.py
```diff
@@ -728,6 +979,7 @@ def default(self):
                 DEFAULTSDIR, 'SAM/WY Southern-Flat Lands.csv')
             self._default = pysam_wind.default('WindPowerNone')
             self._default.WindResourceFile.wind_resource_filename = res_file
+            # self._default.wind_resource_filename = res_file
```
grantbuster: Remove!

reV/generation/generation.py
```diff
@@ -97,9 +103,105 @@ class Gen:
                  'lcoe_nom': {'scale_factor': 1, 'units': 'dol/MWh',
                               'dtype': 'float32', 'chunks': None,
                               'type': 'scalar'},
-                 'lcoe_fcr': {'scale_factor': 1, 'units': 'dol/MWh',
+                 'lcoe_fcr': {'scale_factor': 1, 'units': 'dol/mwh',
```
grantbuster: Why? MW should be capitalized.

reV/handlers/sam_resource.py
```diff
@@ -83,6 +115,9 @@ def __init__(self, project_points, time_index, require_wind_dir=False):
             if not require_wind_dir:
                 self._res_arrays['winddirection'] = np.zeros(self._shape,
                                                              dtype='float32')
+        else:
+            raise HandlerValueError(f'Selected tech {project_points.tech} is'
```
grantbuster: you should set the error message to a string, do logger.error(e) and also raise the exception.

reV/handlers/sam_resource.py
```diff
@@ -281,13 +317,14 @@ def check_units(var_name, var_array, tech):
...
 
-        elif 'pressure' in var_name and tech.lower() == 'csp':
-            if np.min(var_array) < 200:
-                # convert pressure from 100 to 1000 hPa
-                var_array *= 10
-            if np.min(var_array) > 80000:
-                # convert pressure from Pa to hPa
-                var_array /= 100
+        elif 'pressure' in var_name:
+            if tech.lower() == 'csp' or tech.lower() == 'troughphysicalheat':
```
grantbuster: Why only for trough? Why not SWH and Linear as well? Could make a list of these and do `if tech.lower() in list:`

reV/handlers/sam_resource.py
```diff
@@ -360,20 +397,9 @@ def _check_physical_ranges(self, var, arr, var_slice):
...
-        elif self._tech == 'pv':
-            if var in self.PV_DATA_RANGES:
-                valid_range = self.PV_DATA_RANGES[var]
-                arr = self.enforce_arr_range(var, arr, valid_range, arr_sites)
-
-        elif self._tech == 'csp':
-            if var in self.CSP_DATA_RANGES:
-                valid_range = self.CSP_DATA_RANGES[var]
-                arr = self.enforce_arr_range(var, arr, valid_range, arr_sites)
+        if var in self.DATA_RANGES[self._tech]:
```
grantbuster: Nice!

tests/test_gen_swh.py
```diff
@@ -0,0 +1,78 @@
...
+                      'beam', 'diffuse', 'I_incident', 'I_transmitted',
+                      'annual_Q_deliv', 'Q_deliv', 'cf_mean', 'solar_fraction')
+
+    # run reV 2.0 generation
+    gen = Gen.reV_run(tech='solarwaterheat', points=points,
+                      sam_files=sam_files, res_file=res_file, max_workers=1,
+                      output_request=output_request,
+                      sites_per_worker=1, fout=None, scale_outputs=True)
+
+    #  for var in output_request:
```
grantbuster: remove

tests/test_gen_linear.py
```diff
@@ -0,0 +1,82 @@
...
+import numpy as np
+
+from reV.generation.generation import Gen
+from reV import TESTDATADIR
+
+
+def test_gen_linear(caplog):
+    """Test generation for linear Fresnel"""
+
+    caplog.set_level(logging.DEBUG)
```
grantbuster: what is caplog? Typically we don't actively log in tests.

reV/generation/generation.py
```diff
@@ -97,9 +103,105 @@ class Gen:
...
                               'dtype': 'float32', 'chunks': None,
                               'type': 'scalar'},
+                 # Solar water heater
+                 'T_amb': {'scale_factor': 1, 'units': 'C',
```
grantbuster: This is getting to be a lot of outputs! Which is good. But it's getting cumbersome. @MRossol, do you think we should put this in a json? If we put it in the same dir as this file, we can access it using os.path.dirname(__file__), right?

MRossol: I like that idea. We could add an outputs.json to reV/generation and reV/econ and handles the output dictionaries and load them into the module.

If we do that I'd also suggest that we set the path as a cls variable or kwarg which would allow users to provide a custom json (unless that will break $hit)

reV/SAM/SAM.py
mikebannis: gone

reV/SAM/generation.py
mikebannis: fixed

reV/SAM/generation.py
mikebannis: Removed.

reV/SAM/generation.py
mikebannis: Changing it to True causes SAM to crash. Still working on this

reV/SAM/generation.py
mikebannis: Done

reV/SAM/generation.py
mikebannis: Gone

reV/SAM/generation.py
mikebannis: Updated. SAM ignores Feb 29th date in csv files. For drop_leap=False, Dec 31st data is dropped, and Feb 29th is relabeled Mar 1st, March 1st relabeled March 2nd, etc.

reV/handlers/sam_resource.py
mikebannis: Done

reV/handlers/sam_resource.py
mikebannis: Done

tests/test_gen_swh.py
mikebannis: Done

reV/handlers/sam_resource.py
mikebannis: Thank you.

tests/test_gen_linear.py
mikebannis: Removed

reV/generation/generation.py
MRossol: @grantbuster  and @mikebannis  do we want to implement this here or in a new PR?

mikebannis: I would say new PR. The current changes need to get pulled into reV for Parthiv

reV/SAM/SAM.py
```diff
@@ -536,19 +539,15 @@ def ensure_res_len(res_arr, base=8760):
...
             raise ResourceError(msg)
 
         if len(res_arr) % base != 0:
-            div = np.floor(len(res_arr) / 8760)
-            target_len = int(div * 8760)
-            warn('Resource array length is {}, but SAM requires a multiple of '
-                 '8760. Truncating the timeseries to length {}.'
-                 .format(len(res_arr), target_len), SAMInputWarning)
+            div = np.floor(len(res_arr) / base)
+            target_len = int(div * base)
```
grantbuster: Thanks!

reV/SAM/generation.py
```diff
@@ -624,6 +625,337 @@ def default(self):
...
+        m.to_csv(fname, index=False, mode='w')
+
+        # --------- Process data
+        df = resource.copy()
+        df['dt'] = df.index
+
+        # Adjust from UTC to local time
+        rolled = np.roll(df.to_numpy(), timezone * self.time_interval,
+                         axis=0)
+        df = pd.DataFrame(rolled, columns=df.columns, index=df.index)
```
grantbuster: Okay so rolling everything together here is good, but what I would do is make DF as a blank dataframe with index as the resource.index and always drop the leap day if present. Then you use ensure_res_len() when you're setting 1D arrays to the columns in DF.  I think that would be a cleaner way to handle leap days. The stuff you've got going on down at 752 is super confusing

reV/SAM/generation.py
```diff
@@ -624,6 +625,337 @@ def default(self):
...
+
+        Parameters
+        ----------
+        delete_wfile : bool
+            Delete PySAM weather file after processing is complete
+        """
+        super()._gen_exec()
+
+        if delete_wfile:
+            logger.debug('Removing PySAM weather file')
```
grantbuster: probably unnecessary log message.

reV/generation/generation.py
```diff
@@ -100,6 +106,102 @@ class Gen:
                  'lcoe_fcr': {'scale_factor': 1, 'units': 'dol/MWh',
                               'dtype': 'float32', 'chunks': None,
                               'type': 'scalar'},
+                 # Solar water heater
+                 'T_amb': {'scale_factor': 1, 'units': 'C',
+                           'dtype': 'int16', 'chunks': None,
+                           'type': 'array'},
+                 'T_cold': {'scale_factor': 1, 'units': 'C',
+                            'dtype': 'float32', 'chunks': None,
+                            'type': 'array'},
```
grantbuster: One important thing that I didnt catch before. We always want to output profile data in UTC to match the input resource data. So we roll everything to local time for SAM to do the solar position stuff, then roll it back to UTC for the output data. You can see us doing that in `gen_profile()` below. And you can also see how we intercept data in `collect_outputs()` to make sure this happens. Can you think of a way to intercept all arrays coming out of your new solar thermal classes and roll to UTC? Or actually for all generation classes? 

https://github.com/NREL/reV/blob/6964f5f2d660e6877a51158ee75557549aec7d09/reV/SAM/generation.py#L180-L195
https://github.com/NREL/reV/blob/6964f5f2d660e6877a51158ee75557549aec7d09/reV/SAM/generation.py#L197-L215

reV/SAM/generation.py
mikebannis: gone.

mikebannis:
Rolling back output data to UTC is now performed automatically for appropriate outputs. A couple notes:

- The resource time stamps are being saved. Ideally this would have occurred at `Generation.__init__()`. However resource data is not passed to the super classes of `Solar` and `Wind`, hence, the time stamps are being saved in the constructors for `Solar` and `Wind`. This is resulting in several pylint disables at:     https://github.com/NREL/reV/blob/b1dd3eb1247fb8652b41c9c79a184eff0c15783e/reV/SAM/generation.py#L232-L244 Any thoughts on a better way to handle this?

- Outputs that are being rolled back are converted to ndarrays w/o specifying the dtype, which defaults to either int64 or float64. We only need uint16 or float32 at most. Is the memory usage acceptable?

- The existing time zone rolling code casts the roll "step" to int. Is it possible for it to be a float? Would it be better to throw an exception if it was a float versus silently converting to int? https://github.com/NREL/reV/blob/b1dd3eb1247fb8652b41c9c79a184eff0c15783e/reV/SAM/generation.py#L1020-L1026

- I am unable to run test_gen_wind on my computer. @grantbuster, could verify wind is working?

PR #140, opened 2020-03-13
Bug fix on transmission table parse

grantbuster:
So I think there's actually a bug in the CAN transmission data, where an SC point will connect to a substation but not that one line belonging to the substation. The reV SC module merges the SC and Transmission tables and would previously drop any transmission features that doesnt connect to one of the active SC points (a line belonging to an important substation was getting dropped). Now it maintains ALL of the transmission features and drops those that don't connect to an SC point at the very end before the SC sorted buildout.

### DIFF
--- a/reV/handlers/transmission.py
+++ b/reV/handlers/transmission.py
@@ -239,6 +239,7 @@ class TransmissionFeatures:
                     if gid not in missing:
                         missing[gid] = []
                     missing[gid].append(line_gid)
+
         if any(missing):
             emsg = ('Transmission feature table has {} parent features that '
                     'depend on missing lines. Missing dependencies: {}'
--- a/reV/handlers/transmission.py
+++ b/reV/handlers/transmission.py
@@ -48,6 +48,7 @@ class TransmissionFeatures:
             attached lines, legacy method
         """
 
+        logger.debug('Trans table input: {}'.format(trans_table))
         logger.debug('Line tie in cost: {} $/MW'.format(line_tie_in_cost))
         logger.debug('Line cost: {} $/MW-mile'.format(line_cost))
         logger.debug('Station tie in cost: {} $/MW'
@@ -68,6 +69,7 @@ class TransmissionFeatures:
         self._available_capacity_frac = available_capacity
 
         self._features = self._get_features(trans_table)
+        self._check_feature_dependencies()
 
         self._feature_gid_list = list(self._features.keys())
         self._available_mask = np.ones(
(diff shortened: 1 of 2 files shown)


reV/handlers/transmission.py
```diff
@@ -228,6 +230,22 @@ def _get_features(self, trans_table):
...
+    def _check_feature_dependencies(self):
+        """Check features for dependencies that are missing and raise error."""
+        missing = {}
+        for gid, feature_dict in self._features.items():
+            for line_gid in feature_dict.get('lines', []):
+                if line_gid not in self._features:
+                    if gid not in missing:
+                        missing[gid] = []
+                    missing[gid].append(line_gid)
+        if any(missing):
```
MRossol: add line break

reV/handlers/transmission.py
```diff
@@ -228,6 +230,22 @@ def _get_features(self, trans_table):
 
         return features
 
+    def _check_feature_dependencies(self):
```
MRossol: Nice! Damn garbage in ;-)

reV/supply_curve/supply_curve.py
```diff
@@ -70,7 +74,10 @@ def __init__(self, sc_points, trans_table, fcr, sc_features=None,
         self._consider_friction = consider_friction
         self._calculate_total_lcoe_friction()
 
+        mask = ~pd.isna(self._trans_table['sc_gid'])
```
MRossol: Turn this into a static method that would return trans_table, sc_gids, and mask

reV/supply_curve/supply_curve.py
```diff
@@ -284,7 +291,8 @@ def _compute_lcot(trans_table, fcr, trans_costs=None, max_workers=None,
             feature = TC(trans_table, line_limited=line_limited,
                          **trans_costs)
             cost = []
-            for _, row in trans_table.iterrows():
+            iterrows = trans_table[~pd.isna(trans_table['sc_gid'])].iterrows()
```
MRossol: Since your already defining `mask = ~pd.isna(trans_table['sc_gid'])` below, why not have this be:
```
mask = ~pd.isna(trans_table['sc_gid'])
for _, row in  trans_table[mask].iterrows():
```

reV/supply_curve/supply_curve.py
```diff
@@ -413,6 +425,17 @@ def _parse_trans_table(sc_points, trans_table, fcr, trans_costs=None,
                    .format(len(missing), missing))
             logger.warning(msg)
             warn(msg)
```
MRossol: add line break

reV/supply_curve/supply_curve.py
```diff
@@ -399,8 +409,10 @@ def _parse_trans_table(sc_points, trans_table, fcr, trans_costs=None,
         rename = {p: t for p, t in zip(point_merge_cols, table_merge_cols)}
         sc_cap = sc_cap.rename(columns=rename)
 
+        logger.debug('Merging SC table and Trans Table on columns: {}'
+                     .format(table_merge_cols))
         trans_table = trans_table.merge(sc_cap, on=table_merge_cols,
```
MRossol: This method has become pretty long and complex. What do you think about pulling out this middle bit that does the merge between sc_points and the trans_table and putting that into a new staticmethod?

grantbuster: Agreed! Broke out all methods into single purpose static methods (merge, check, lcot). This is the problem with multi-purpose "parse()" methods.... They become do-everything methods. Gotta be cautious of that in the future.

reV/supply_curve/supply_curve.py
grantbuster: Agreed, moved the `gid_mask` up to the top of the method

reV/supply_curve/supply_curve.py
grantbuster: Agreed!

reV/handlers/transmission.py
grantbuster: Seriously..... The CONUS trans table was solid, no bugs, but i think the CAN table was missing stuff. This helps.

PR #143, opened 2020-03-20
Gb/append econ

grantbuster:
Added an append option to the econ module to append data to the generation input h5 files. Should work great when needing to run gen -> econ as separate modules in the reV pipeline.

### DIFF
--- a/reV/config/sam_analysis_configs.py
+++ b/reV/config/sam_analysis_configs.py
@@ -261,7 +261,7 @@ class EconConfig(SAMAnalysisConfig):
                 # need to make list of res files for each year
                 self._cf_files = [fname.format(year) for year in self.years]
             elif 'PIPELINE' in fname:
-                self._cf_files = Pipeline.parse_previous(self.dirout,
+                self._cf_files = Pipeline.parse_previous(super().dirout,
                                                          'econ',
                                                          target='fpath')
             else:
--- a/reV/config/sam_analysis_configs.py
+++ b/reV/config/sam_analysis_configs.py
@@ -6,6 +6,7 @@ Created on Mon Jan 28 11:43:27 2019
 
 @author: gbuster
 """
+import os
 import logging
 from math import ceil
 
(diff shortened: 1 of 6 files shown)


reV/config/sam_analysis_configs.py
```diff
@@ -296,3 +297,29 @@ def site_data(self):
...
+
+        Returns
+        -------
+        dirout : str
+            Target path for reV output files.
+        """
+        self._dirout = super().dirout
+        if self.append:
+            self._dirout = os.path.dirname(self.cf_files[0])
+        return self._dirout
```
MRossol: add line break

reV/econ/econ.py
```diff
@@ -141,9 +144,13 @@ def __init__(self, points_control, cf_file, cf_year, site_data=None,
...
 
-        # initialize output file
-        self._init_fpath()
-        self._init_h5()
+        # initialize output file or append econ data to gen file
+        if append:
+            self._fpath = self._cf_file
+        else:
+            self._init_fpath()
+        mode = 'a' if append else 'w'
```
MRossol: add line break

reV/econ/econ.py
```diff
@@ -141,9 +144,13 @@ def __init__(self, points_control, cf_file, cf_year, site_data=None,
...
+        self._init_h5(mode=mode)
```
MRossol: I don't understand how this doesn't break? Output.init_h5 will try to create all of the datasets which will break if they already exist. Am I missing something?

reV/handlers/outputs.py
```diff
@@ -522,8 +522,13 @@ def _create_dset(self, ds_name, shape, dtype, chunks=None, attrs=None,
             Dataset attributes
         data : ndarray
             Dataset data array
+        replace : bool
+            If previous dataset exists with the same name, it will be replaced.
         """
         if self.writable:
+            if ds_name in self.datasets and replace:
+                del self.h5[ds_name]
+
```
MRossol: need an elif that skips the create_dataset step.

grantbuster: Previous behavior was to break on h5py "link already exists" error. I think this is important if the shape and dtype don't match. I'm adding a new commit that checks for this.

reV/econ/econ.py
grantbuster: Econ should only be making new datasets, plus I added better handling of pre-existing datasets.

grantbuster:
@MRossol, approve? Travis used this last Friday and the results look good. I think its an important feature.

reV/handlers/outputs.py
MRossol: I'm surprised this logic works, the else won't be reached after you delete the dset so you won't re-create the dataset

grantbuster: Yep you're right. fixed.

grantbuster:
No problem! It's improved now. 

I dont think we need a new release? Its a new "feature" but super super minor.

grantbuster:
@MRossol, actually, since Mike added in those solar thermal models I think it might be worth doing a feature release and might as well include this too.

PR #148, opened 2020-04-01
Fixed a bug where categorical data aggregation gets set in dataframe …

grantbuster:
…as a dict with int/float keys and PD doesnt save to CSV in machine readable format. Forced json dump in method.

### DIFF
--- a/reV/supply_curve/point_summary.py
+++ b/reV/supply_curve/point_summary.py
@@ -5,6 +5,7 @@ Created on Fri Jun 21 13:24:31 2019
 
 @author: gbuster
 """
+import json
 import logging
 import numpy as np
 import pandas as pd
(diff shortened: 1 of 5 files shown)


reV/supply_curve/point_summary.py
```diff
@@ -475,17 +476,58 @@ def _agg_data_layer_method(data, excl_mult, method):
...
                 e = ('Cannot recognize data layer agg method: '
-                     '"{}". Can only do mean, mode, sum, or category_count.'
+                     '"{}". Can only do mean, mode, sum, or category.'
                      .format(method))
                 logger.error(e)
                 raise ValueError(e)
 
         return data
 
+    def _agg_data_layer_category(data, excl_mult):
```
MRossol, line 490: What about having this class return the dictionary, and then add a staticmethod to jump to str that you'd run when you're adding the data to the DataFrame. Long story short, I think it would be nice to be able to get the original dict

PR #150, opened 2020-04-10
Misc wind resource bug fixes: enforced tech name to be windpower, better

grantbuster:
resource limits on pressure, better identification of unit ranges with
median.

### DIFF
--- a/reV/SAM/generation.py
+++ b/reV/SAM/generation.py
@@ -977,6 +977,7 @@ class Wind(Generation):
         var_list = ['temperature', 'pressure', 'windspeed', 'winddirection']
         if 'winddirection' not in resource:
             resource['winddirection'] = 0.0
+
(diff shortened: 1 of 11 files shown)


reV/SAM/generation.py
```diff
@@ -975,6 +975,8 @@ def set_wtk(self, resource):
 
         data_dict = {}
         var_list = ['temperature', 'pressure', 'windspeed', 'winddirection']
+        if 'winddirection' not in resource:
+            resource['winddirection'] = 0.0
         time_index = resource.index
```
MRossol, line 980: add line break

grantbuster:
offshore is not run any different from the SAM perspective (and never was). Offshore is only delineated when we run the offshore module (runs aggregation of gen points to wind farms and also runs ORCA).

PR #155, opened 2020-04-17
Rex refactor

MRossol:
Migrate resource handlers and utilities to rex


reV/handlers/__init__.py
```diff
@@ -3,7 +3,6 @@
 Sub-package of data handlers
 """
 from .collection import Collector
+from .exclusions import ExclusionLayers
 from .multi_year import MultiYear
-from .resource import Resource
-from .rev_resource import (NSRDB, MultiFileNSRDB, MultiFileWTK,
-                           SolarResource, WindResource)
+from .outputs import Outputs
```
grantbuster, line 8: Eventually we might want to think about moving collection, multi year, and outputs to rex as well. I'm sure a lot of people in SEAC would start using those tools if they were flexible enough. For now though reV is an easy dependency too.

tests/peregrine.py
```diff
@@ -19,8 +19,8 @@
 
 from reV import TESTDATADIR
 from reV.handlers.outputs import Outputs
-from reV.utilities.execution import PBS
-from reV.utilities.loggers import init_logger
+from rex.utilities.execution import PBS
```
grantbuster, line 22: We might just want to remove this file (peregrine.py). We're definitely not maintaining it.

reV/handlers/__init__.py
MRossol: I could see outputs over Collection and MY, but it still feels like more of a reV tool than a rex utility, at least we now have the option for later!

tests/peregrine.py
MRossol: love it, gone!

Issue #159, opened 2020-04-22
Wind Curtailment Affecting ws_mean Output


grantbuster:
**Bug Description**
Wind curtailment filters the wind speed data before being input to SAM as a way to set generation during curtailed timesteps to zero. This affects the calculation of ws_mean (mean windspeed), which can also impact supply curve tables with resource binning. 

**Expected behavior**
ws_mean should probably be the same between curtailed and non-curtailed output since the actual wind speed is not what is being curtailed (in real life).

**Additional context**
The resource object is manipulated [here](https://github.com/NREL/reV/blob/791a9fd453d74e4e6e8c452bce7c3e8df3b89acb/reV/SAM/generation.py#L308) for wind curtailment, and the ws_mean is calculated afterwards from the modified resource data [here](https://github.com/NREL/reV/blob/791a9fd453d74e4e6e8c452bce7c3e8df3b89acb/reV/SAM/generation.py#L322). 

Two obvious options would be:
1) Make a copy of the resource data before curtailment and use that to calculate resource means. This would be data inefficient. 
2) Have the curtailment method output a curtailment mask which gets used to manipulate the SAM outputs [after this point](https://github.com/NREL/reV/blob/791a9fd453d74e4e6e8c452bce7c3e8df3b89acb/reV/SAM/generation.py#L332). This would have the disadvantage of having to parse the SAM outputs explicitly, and we might miss some stuff like monthly outputs that are uncommon,

A suspected short-term workaround is to run reV supply curve without resource bins.

MRossol:
fixed in #162

PR #162, opened 2020-04-30
Mean res refactor

MRossol:
Fix for #159 by pre-computing mean_res in SAMResource and then extracting it

### DIFF
--- a/reV/version.py
+++ b/reV/version.py
@@ -1,3 +1,3 @@
 """reV Version number"""
 
-__version__ = "0.3.1"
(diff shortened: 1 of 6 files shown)


reV/version.py
```diff
@@ -1,3 +1,5 @@
-"""reV Version number"""
+"""
+reV Version number
```
grantbuster, line 2: You know you can rerun all checks on github right?

MRossol: NO!

PR #163, opened 2020-05-14
QA/QC

MRossol:
Add QA/QC module with:
Summarize:
- Class to summarize datasets using pandas.describe
SummaryPlots:
- Plot scatter, supply_curve, and dist plots of summary data


MRossol:
@grantbuster Things to think about:
1) What do we want to be the default plots: 
- .pngs from plotting
- .html from plotly
2) Do we want to make classes for QA on each module? If so maybe qa_qc becomes a sub_package instead of a utility?
3) Besides summary tables and plots what other QA should we do?

grantbuster:
@MRossol, looks good but what about 1) summary statistics for csv files (supply curve mean and total capacities, lcoe, etc...) 2) integration into pipeline?

MRossol:
@grantbuster  for SC summary, should I compute the mean, stdev, median, min, max, and sum for all numerical columns? Or a subset of columns? 

I'm working on a QA/QC CLI, but would love your thoughts on pipeline integration...

grantbuster:
Would be really quick to do all columns right? Any reasons why we wouldn't want to do all numeric columns?

Want to set up a call on Monday about pipeline integration?

MRossol:
Agreed and sure!

MRossol:
@grantbuster QA/QC CLI is updated with config option. I'll start working on a config example but would love your help integrating with the pipeline and updating the pipeline example

grantbuster:
@MRossol, sure thing! Can you send me the config example when its done? Or just add it to the current pipeline example? I'll try and get to that this week if thats okay.

MRossol:
@grantbuster qa_qc config example added to the pipeline example folder

PR #166, opened 2020-05-21
Gb/qa pipeline

grantbuster:


reV/qa_qc/cli_qa_qc.py
```diff
@@ -244,15 +262,18 @@ def from_config(ctx, config_file, verbose):
...
+                config.dirout, 'qa-qc', name, replace=True,
                 job_attrs={'hardware': 'local',
                            'dirout': config.dirout})
 
-            for module in config.module_names:
+            terminal = False
+            for i, module in enumerate(config.module_names):
+                if i == len(config.module_names) - 1:
+                    terminal = True
                 module_config = config.get_module_inputs(module)
```
MRossol, line 277: line break

reV/qa_qc/summary.py
```diff
@@ -191,9 +191,11 @@ def summarize_dset(self, ds_name, process_size=None, max_workers=None,
             ds_shape, _, ds_chunks = f.get_dset_properties(ds_name)
             if len(ds_shape) > 1:
                 sites = np.arange(ds_shape[1])
-                if max_workers > 1:
+                if max_workers != 1:
+                    if process_size is None and ds_chunks is not None:
```
grantbuster, line 195: @MRossol, i think this is what you were trying to do? The code ran into some issues i think when ds_chunks was none.

reV/qa_qc/cli_qa_qc.py
```diff
@@ -321,23 +344,25 @@ def get_h5_cmd(name, h5_file, out_dir, dsets, group, process_size, max_workers,
 
     if verbose:
         args += '-v '
+    if terminal:
```
MRossol, line 347: line break

reV/qa_qc/cli_qa_qc.py
```diff
@@ -350,14 +375,16 @@ def get_sc_cmd(name, sc_table, out_dir, columns, plot_type, lcoe, log_file,
...
```
MRossol, line 378: line break

reV/qa_qc/summary.py
```diff
@@ -191,9 +191,11 @@ def summarize_dset(self, ds_name, process_size=None, max_workers=None,
...
+                        process_size = ds_chunks[1]
                     if process_size is None:
```
MRossol, line 197: line break

grantbuster:
Line breaks added. Regarding the terminal flag: the way we've set up the pipeline is that it looks for job status json files in the output directory to signal that a module is done. because we're doing multiple CLI calls, we don't want to print a status json with a successful flag until the last cli call. So the terminal flag says "this is the terminal cli call, print the status json when complete". It's not perfect - if an intermediate cli call fails the status json could still show successful.

PR #168, opened 2020-05-22
Qa excl

MRossol:
Add heatmap plots of final exclusions mask

Full pipeline test is in progress

### DIFF
--- a/docs/source/reV/reV.config.qa_qc_config.rst
+++ b/docs/source/reV/reV.config.qa_qc_config.rst
@@ -1,0 +1,7 @@
+reV.config.qa\_qc\_config module
+================================
+
+.. automodule:: reV.config.qa_qc_config
+   :members:
+   :undoc-members:
+   :show-inheritance:
--- a/docs/source/reV/reV.config.rst
+++ b/docs/source/reV/reV.config.rst
@@ -22,6 +22,7 @@ Submodules
    reV.config.output_request
    reV.config.pipeline
    reV.config.project_points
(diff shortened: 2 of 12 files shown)


MRossol:
Full test ran, no wonder if died, the .npy file is 3.1 GB!

MRossol:
Okay, @grantbuster  let me know what you think. I heavily refactored summary, but only lightly changed QaQc. I kept just one QaQc class but reduced in init to just be out_dir, so its mainly just a set of staticmethods and classmethods to run QaQc on h5, sc, or excl

grantbuster:
@MRossol, looks great. I agree its maybe a bit cumbersome with the multiple classes but the heirarchy is so much more obvious. 

Any reason to run the test on Eagle again? Can you update the example pipeline config? Maybe add in the exclusion plot interval kwarg?

MRossol:
I just re-ran the cli test and no issues. The only real excl kwarg would be point_step... I can change that to 200 or 1000 for the test?

grantbuster:
@MRossol, yeah i more just think its nice to have in the example so anyone copying the example config knows its an options. 

Can you add the exclusion block to the example file though? I just made a smaller standalone config for just the exclusions on Eagle and it never made its way to this file: https://github.com/NREL/reV/blob/qa_excl/examples/full_pipeline_execution/config_qa-qc.json

MRossol:
I just came to the same conclusion when I went back to look at the example config. Just added the excl module to the example config from your eagle file

Issue #172, opened 2020-05-28
Percent inclusions should just use the minimum inclusion


nickwg03:
**Bug Description**
With overlapping percent inclusions, we don't want to multiply a single pixel repeatedly with progressively smaller available area. For all overlapping percent inclusions, we just want the minimum inclusion. For example, with four overlapping layers of 0.25, 0.50, 0.75, and 1.00 inclusion, we'd just want to use the 0.25 inclusion for that pixel.

**Charge code**
**WCRD.11447.01.01.01**

MRossol:
Implemented in #173

PR #173, opened 2020-05-28
change exclusions to element-wise-minima

MRossol:
fix for #172

### DIFF
--- a/tests/test_supply_curve_exclusions.py
+++ b/tests/test_supply_curve_exclusions.py
@@ -248,6 +248,32 @@ def test_no_excl(ds_slice):
         assert np.allclose(truth, test)
 
 
+def test_multiple_excl_fractions():
+    """
+    Test that multiple fraction exclusions are handled properly
+    """
+    excl_h5 = os.path.join(TESTDATADIR, 'ri_exclusions', 'ri_exclusions.h5')
+
+    excl_dict = {'ri_padus': {'exclude_values': [1, ], 'weight': 0.5,
+                              'exclude_nodata': True}}
+    with ExclusionMaskFromDict(excl_h5, layers_dict=excl_dict) as f:
+        truth = f.mask
+
+    excl_dict = {'ri_padus': {'exclude_values': [1, ], 'weight': 0.25,
+                              'exclude_nodata': True}}
+    with ExclusionMaskFromDict(excl_h5, layers_dict=excl_dict) as f:
+        truth = np.minimum(truth, f.mask)
+
+    excl_dict = {'ri_padus': {'exclude_values': [1, ], 'weight': 0.25,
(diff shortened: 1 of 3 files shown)


tests/test_supply_curve_exclusions.py
```diff
@@ -248,6 +248,32 @@ def test_no_excl(ds_slice):
...
+        truth = np.minimum(truth, f.mask)
+
+    excl_dict = {'ri_padus': {'exclude_values': [1, ], 'weight': 0.25,
+                              'exclude_nodata': True},
+                 'ri_padus': {'exclude_values': [1, ], 'weight': 0.25,
+                              'exclude_nodata': True}}
+    with ExclusionMaskFromDict(excl_h5, layers_dict=excl_dict) as f:
+        test = f.mask
+
+    assert np.allclose(test, truth)
```
grantbuster, line 274: Would you also want to add a line checking that the minimum value greater than zero is not less than 0.25? Simple check that would be worth doing, imo.

tests/test_supply_curve_exclusions.py
```diff
@@ -248,6 +248,32 @@ def test_no_excl(ds_slice):
...
+                              'exclude_nodata': True}}
+    with ExclusionMaskFromDict(excl_h5, layers_dict=excl_dict) as f:
+        truth = f.mask
+
+    excl_dict = {'ri_padus': {'exclude_values': [1, ], 'weight': 0.25,
+                              'exclude_nodata': True}}
+    with ExclusionMaskFromDict(excl_h5, layers_dict=excl_dict) as f:
+        truth = np.minimum(truth, f.mask)
+
+    excl_dict = {'ri_padus': {'exclude_values': [1, ], 'weight': 0.25,
```
grantbuster, line 267: Should this be `'weight': 0.5`?

tests/test_supply_curve_exclusions.py
nickwg03: @grantbuster we'd potentially have percent inclusions that are less than 25%, though. Could be any number between 0 and 1, really.

grantbuster: @nickwg03, this is a pytest. This would be just a verification on the exclusion data based on the test inputs. I'm not implying anything about production outputs.

nickwg03: Haha, okay. I'll stay in my lane.

grantbuster: Lol sorry if that was blunt, I appreciate you reviewing our PR's :)

tests/test_supply_curve_exclusions.py
MRossol: yup, will fix,

PR #184, opened 2020-07-13
config QA

MRossol:
Add _check_keys to configs
Add option to overwrite entries
Add valid_keys to CLI as subcommand


MRossol:
@grantbuster FYI full test passed after fixing a few more bugs.

I also started a branch on reVX and ended up making the REQUIREMENTS preflight part of BaseConfig which cleaned things up nicely

PR #185, opened 2020-07-27
ProjectPoint entry points for lat, lon coords or geographic regions

MRossol:


reV/config/cli_project_points.py
```diff
@@ -0,0 +1,190 @@
...
+@click.group()
+@click.option('--fpath', '-f', type=click.Path(), required=True,
+              help='.csv file path to save project points to')
+@click.option('--res_file', '-rf', required=True,
+              help=('Filepath to single resource file, multi-h5 directory, '
+                    'or /h5_dir/prefix*suffix.'))
+@click.option('--tech', '-t', type=STR, default=None,
+              help='reV tech to analyze (required).')
+@click.option('--sam_files', '-sf', required=True, type=SAMFILES,
+              help='SAM config files (required) (str, dict, or list).')
```
grantbuster, line 117: Why are sam configs required to make a project points csv file? Is there any mapping done between points and multiple sam files? What is put in the config column of the project points csv? Maybe ill find out later...

reV/config/cli_project_points.py
```diff
@@ -0,0 +1,190 @@
...
+@click.option('--curtailment', '-curt', type=click.Path(exists=True),
+              default=None,
+              help=('JSON file with curtailment inputs parameters. '
+                    'Default is None (no curtailment).'))
```
grantbuster, line 121: Why is this needed to make a project points csv file? Maybe ill found out later...

reV/config/project_points.py
```diff
@@ -721,3 +724,160 @@ def split(cls, i0, i1, project_points):
...
+        ----------
+        lat_lons : str | tuple | list | ndarray
+            Pair or pairs of latitude longitude coordinates
+        res_file : str
+            Resource file, needed to fine nearest neighbors
+        sam_config : dict | str | list | SAMConfig
+            SAM input configuration ID(s) and file path(s). Keys are the SAM
+            config ID(s), top level value is the SAM path. Can also be a single
+            config file str. If it's a list, it is mapped to the sorted list
+            of unique configs requested by points csv. Can also be a
```
grantbuster, line 769: I dont think this doc string is applicable to this class method. What would happen if you input a list for this? Same for the next class method

MRossol:
@grantbuster  I made a change to [reV.generation.Gen.out](https://github.com/NREL/reV/blob/0f9c00072f34a10c65c21a34e2a5818a50bebe7a/reV/generation/generation.py#L1040-L1062) so that the property returns the UNSCALED (or native PySAM) values instead of the scaled .h5 native values. This makes it a lot more user friendly when running with fout=None.

Running tests now to make sure I changed all internal call in flush to self._out.

Wanted to get your approval as well.

grantbuster:
@MRossol yep makes sense as long as the internal linkages still use the scaled self._out attr.

PR #192, opened 2020-08-11
update offshore sc point gids

MRossol:
Map offshore wind Farms to nearest sc_point_gids
Make offshore transmission table optional and map wind farm gids to sc_point_gids in sc_agg table
Update CLI and config
Expose offshore_wind_dirs option in CLI and config


MRossol:
@grantbuster do you mind doing a quick review? It was pretty straight forward along with some cleanup along the way.

grantbuster:
@MRossol yep will do!

reV/config/supply_curve_configs.py
```diff
@@ -293,6 +301,13 @@ def downwind(self):
         """Get the flag to exclude downwind as well as upwind sites"""
         return self.get('downwind', False)
 
+    @property
+    def offshore_wind_dirs(self):
+        """
+        Get flag to execute competitive wind directions for offshore sites
```
grantbuster, line 307: Is this really a flag? Or was this docstring not updated? "dirs" seems to indicate it should be a points to wind direction data? Does it also work as a flag since the default is False?

After reviewing the CLI, it looks like this really is a flag. Maybe we can rename it to differentiate from the wind_dirs input?

reV/handlers/exclusions.py
```diff
@@ -144,8 +144,7 @@ def layers(self):
         layers : list
             List of exclusion layers
         """
-        layers = [ds for ds in self.h5
-                  if ds != 'meta']
```
grantbuster, line 148: No meta data in exclusion files? Do we not have to filter out the lat/lon datasets?

reV/offshore/offshore.py
```diff
@@ -153,14 +153,15 @@ def _parse_cf_meta(gen_fpath):
...
             meta = out.meta
+
         if 'offshore' not in meta:
             e = ('Offshore module cannot run without "offshore" flag in meta '
                  'data of gen_fpath: {}'.format(gen_fpath))
             logger.error(e)
             raise KeyError(e)
 
-        onshore_mask = (meta['offshore'] == 0)
-        offshore_mask = (meta['offshore'] == 1)
```
grantbuster, line 163: This is just a habit of mine. I don't like having more than one equal sign on a line. The parenthesis helps me visually. No worries either way though, just thought i'd note why i do this.

reV/supply_curve/sc_aggregation.py
```diff
@@ -365,11 +365,39 @@ def _parse_meta_cols(offshore_meta_cols, gen_meta):
...
+            Offshore wind farm coordinates (lat, lon) pairs
+
+        Returns
+        -------
+        points : pandas.DataFrame
+            Nearest SC points (gid, row_ind, and col_ind)
+        """
+        with SupplyCurveExtent(excl_fpath, resolution=resolution) as f:
+            points = f.points
+            sc_lat_lons = np.dstack((f.latitude, f.longitude))[0]
```
grantbuster, line 389: Maybe have this 2D lat/lon array as a SCExtent property? Seems useful.

reV/supply_curve/supply_curve.py
```diff
@@ -131,42 +135,253 @@ def _parse_sc_points(sc_points, sc_features=None):
...
+                                        'mean_lcoe')):
+        """Merge the supply curve table with the transmission features table.
+
+        Parameters
+        ----------
+        sc_points : pd.DataFrame
+            Table of supply curve point summary
+        trans_table : pd.DataFrame
+            Table mapping supply curve points to transmission features
+        offshore_table : str, optional
```
grantbuster, line 211: Why is this a str while the other inputs are dataframes?

reV/config/supply_curve_configs.py
MRossol: @grantbuster, good point, I was trying to make it less ambiguous than just offshore, but I guess I failed. Three thoughts:
1) wind_dirs_offshore
2) include_offshore_wind_dirs
3) include_wind_dirs_offshore

reV/handlers/exclusions.py
MRossol: I was surprised as well but makes sense as instead of meta we have a latitude and longitude dataset

reV/offshore/offshore.py
MRossol: But it's not pythonic ;-)

reV/supply_curve/sc_aggregation.py
MRossol: done

reV/supply_curve/supply_curve.py
MRossol: old docstring, fixed

reV/offshore/offshore.py
grantbuster: I'm not sure about that! May i quote the zen of python: beautiful is better than ugly, and readability counts :)

I'll probably keep doing this haha

PR #193, opened 2020-08-12
Gb/nohup

grantbuster:


reV/config/execution.py
```diff
@@ -206,5 +206,5 @@ def walltime(self):
         _hpc_walltime : int
             Requested single node job time in hours.
         """
-        self._hpc_walltime = int(self.get('walltime', self._hpc_walltime))
+        self._hpc_walltime = float(self.get('walltime', self._hpc_walltime))
```
MRossol, line 209: is this what was causing jobs to die if you request < 1hr?

grantbuster: Yep! 0.5 hours got rounded down to 0.0 which slurm didnt like. Annoying little bug!

grantbuster:
@MRossol, agreed, pipeline has two flags now, --monitor and --background (the second only works with the first). Batch only has --monitor-background since you cant actively monitor all batch jobs in the "foreground"

PR #198, opened 2020-08-27
fix meta index name bug

MRossol:
fix meta being named 'gid' during collect

### DIFF
--- a/reV/qa_qc/summary.py
+++ b/reV/qa_qc/summary.py
@@ -207,7 +207,7 @@ class SummarizeH5:
         with Resource(self.h5_file, group=self._group) as f:
             meta = f.meta
             if 'gid' not in meta:
-                if meta.index.name == 'gid':
+                if meta.index.name != 'gid':
                     meta.index.name = 'gid'
 
                 meta = meta.reset_index()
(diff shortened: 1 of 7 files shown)


reV/qa_qc/summary.py
```diff
@@ -207,7 +207,9 @@ def summarize_means(self, out_path=None):
         with Resource(self.h5_file, group=self._group) as f:
             meta = f.meta
             if 'gid' not in meta:
-                meta.index.name = 'gid'
+                if meta.index.name == 'gid':
+                    meta.index.name = 'gid'
```
grantbuster, line 211: seems redundant?

MRossol: whoops, good call, was trying to Not run that unless we need to

grantbuster:
@MRossol can you also patch this while you're at it? This warn message got put in the wrong place somehow. It gets triggered whenever `analysis_years` is input to the config as a list, which is nearly always. Maybe it should just be:
```
            self._years = self.get('analysis_years', [None])
            if not isinstance(self._years, list):
                self._years = [self._years]
            if self._years[0] is None:
                warn()
```

MRossol:
Done

PR #201, opened 2020-09-02
add inclusion_weights and tests

MRossol:


grantbuster (changes requested):
This looks good can you just add a note on the new key to the sc aggregation cli in the help string for excl_dict?

MRossol:
@grantbuster done, I also updated the pipeline example with an inclusion_weights entry which made me realize we need to convert the keys from str to floats

grantbuster (approved):

PR #204, opened 2020-09-08
Gb/eqn curtailment

grantbuster:
Curtailment features:
 - Curtailment timeframe based on dates not months
 - Curtailment boolean based on equation string not thresholds 

Batch features:
 - Don't set folder name from arguments that only have one unique value
 - Make job summary table 
 - Added a --delete flag to remove/clear all batch job directories.


grantbuster:
@MRossol this is good-to-go now. Tested on Eagle and everything works as expected with real job creation and submission.

reV/batch/batch.py
```diff
@@ -64,18 +67,28 @@ def _parse_config(config):
...
         # iterate through batch sets
         for s in config['sets']:
+            set_tag = s.get('set_tag', '')
+            if set_tag in sets:
+                msg = ('Found multiple sets with the same set_tag: "{}"'
+                       .format(set_tag))
+                logger.error(msg)
+                raise ValueError(msg)
+            else:
+                sets.append(set_tag)
```
MRossol, line 91: What happens if set_tag == ''

reV/batch/batch.py
```diff
@@ -88,12 +101,12 @@ def _parse_config(config):
...
                 arg_combs.append(comb_dict)
                 file_sets.append(s['files'])
-                set_tags.append(s.get('set_tag', ''))
+                set_tags.append(set_tag)
 
         return arg_combs, file_sets, set_tags
 
     @staticmethod
-    def _tag_value(value):
+    def _fix_tag_w_year(value):
```
MRossol, line 109: On line 124, doesn't parse_year have a bool option, making the try except redundant?

reV/batch/batch.py
```diff
@@ -167,6 +179,56 @@ def _make_job_tag(self, set_tag, arg_comb):
...
+                ignore_tags.append(arg)
+
+        tag_arg_comb = {k: v for k, v in arg_comb.items()
+                        if k not in ignore_tags}
+
+        return tag_arg_comb
+
+    @property
+    def job_table(self):
+        """Get a dataframe summarizing the batch jobs."""
```
MRossol, line 219: Is there a need to generate this everytime job_table is called? I.e., can it change?

MRossol (commented):
Few minor comments + the usual check import order and line spacing if you have the time

Honestly, I can't follow the minutia but generally, the logic looks good and I trust you.

reV/batch/batch.py
grantbuster: That's the default set tag (no set tag). That's used to partition the permutations and to set up the job names. So the job names look like: `set_tag_{arg_combination_0}_{arg_combination_1}`. I also added a protection so that multiple sets cannot have the same set tag (and multiple sets cannot have undefined set tags): https://github.com/NREL/reV/pull/204/files#diff-6e2feea37f6607cee458554b892246a3R85

reV/batch/batch.py
grantbuster: Oh yeah! Maybe it didnt when i wrote this. Simplified in a new commit.

reV/batch/batch.py
grantbuster: Yeah, this only gets called when all of the batch job directories are made. So if a user has changed the batch permutations the table should be re-created. In any case, I think this is pretty fast compared to all of the other data copying and SLURM job submission that batch does.

MRossol (approved):

PR #209, opened 2020-09-15
bug fix

grantbuster:
bug fix: whenever a sam input dict is passed to a worker it needs to be deepcopied so any modifications (such as tilt at latitude) are not propogated to the top level sam config obj.


grantbuster:
Full conus test running on Eagle now.

MRossol (approved):
Looks good!

PR #211, opened 2020-09-24
Gb/debug

grantbuster:


reV/SAM/econ.py
```diff
@@ -160,57 +160,79 @@ def _get_annual_energy(site, site_df, site_gids, cf_arr, inputs, calc_aey):
...
         with Outputs(cf_file) as cfh:
 
             # get the index location of the site in question
-            site_gids = list(cfh.meta['gid'])
-            isite = site_gids.index(site)
+            site_gids = list(cfh.get_meta_arr('gid'))
+            isites = [site_gids.index(s) for s in sites]
 
             # look for the cf_profile dataset
             if 'cf_profile' in cfh.datasets:
```
MRossol, line 192: What about cleaning this up a bit more to be:

```
            if 'cf_profile' in cfh.datasets:
               dset = 'cf_profile'
            elif 'cf_profile-{}'.format(cf_year) in cfh.datasets:
                dset = 'cf_profile-{}'.format(cf_year)
            elif 'cf_profile_{}'.format(cf_year) in cfh.datasets:
                dset = 'cf_profile_{}'.format(cf_year)
            else:
                msg = ('Could not find cf_profile values for '
                       'input to SingleOwner. Available datasets: {}'
                       .format(cfh.datasets))
                logger.error(msg)
                raise KeyError(msg)

            profiles = cfh[dset, :, isites]
```

reV/SAM/econ.py
```diff
@@ -160,57 +160,79 @@ def _get_annual_energy(site, site_df, site_gids, cf_arr, inputs, calc_aey):
...
+
+        Returns
+        -------
+        inputs : dict
+            Dictionary of SAM input parameters with the generation profile
+            added.
+        """
+
+        sys_cap = Economic._parse_sys_cap(site, inputs, site_df)
+        gen = profiles[:, isite] * sys_cap
```
MRossol, line 235: why not set this directly to the dict, i.e. don't create the intermediate gen variable?

reV/SAM/econ.py
```diff
@@ -160,57 +160,79 @@ def _get_annual_energy(site, site_df, site_gids, cf_arr, inputs, calc_aey):
         return site_df
 
     @staticmethod
-    def _get_gen_profile(site, site_df, cf_file, cf_year, inputs):
-        """Get the single-site generation time series and add to inputs dict.
+    def _get_cf_profiles(sites, cf_file, cf_year):
```
MRossol, line 163: We should really add logic to rex that will check to see if a list can be converted to a slice...

MRossol (commented):
Minor syntax comments, but the logic looks much cleaner!

reV/SAM/econ.py
grantbuster: yes

reV/SAM/econ.py
grantbuster: because thats how it was before hahah. Will change!

reV/SAM/econ.py
grantbuster: Yes but sites may not be consecutive. Going to leave as-is for now.

MRossol (approved):

PR #216, opened 2020-09-28
Gb/agg econ h5 args

grantbuster:


reV/supply_curve/sc_aggregation.py
```diff
@@ -465,6 +465,10 @@ def run(cls, summary, handler, excl_fpath, res_data, res_class_bins,
             data. If this is input as None, this method has been called
             without offshore data in error and summary will be passed
             through un manipulated.
+        h5_dsets_data : dict | None
```
MRossol, line 468: Where does this data come from? Is it extracted on the fly?

MRossol (approved):
Looks good!

PR #218, opened 2020-10-05
ProjectControl cleanup

MRossol:
clean up _pp_to_pc
move append logic to econ

### DIFF
--- a/reV/econ/econ.py
+++ b/reV/econ/econ.py
@@ -323,8 +323,8 @@ class Econ(Gen):
 
         if append and cf_file is not None:
             try:
-                pc = pc._econ_append_pc(pc.project_points, cf_file,
-                                        sites_per_worker=sites_per_worker)
+                pc = Econ._econ_append_pc(pc.project_points, cf_file,
+                                          sites_per_worker=sites_per_worker)
             except Exception:
                 msg = ("Failed to adjust ProjectControl for Econ in append "
                        "mode!")
--- a/reV/econ/econ.py
+++ b/reV/econ/econ.py
@@ -317,9 +317,9 @@ class Econ(Gen):
         pc : reV.config.project_points.PointsControl
             PointsControl object instance.
         """
-        pc = super().get_pc(points, points_range, sam_files, tech,
-                            sites_per_worker=sites_per_worker,
-                            res_file=cf_file)
+        pc = Gen.get_pc(points, points_range, sam_files, tech,
+                        sites_per_worker=sites_per_worker,
+                        res_file=cf_file)
 
         if append and cf_file is not None:
             try:
--- a/reV/econ/econ.py
+++ b/reV/econ/econ.py
@@ -9,13 +9,19 @@ import pandas as pd
 import pprint
 from warnings import warn
 
+from reV.config.project_points import PointsControl
 from reV.generation.generation import Gen
 from reV.handlers.outputs import Outputs
 from reV.SAM.econ import LCOE as SAM_LCOE
 from reV.SAM.econ import SingleOwner
 from reV.SAM.windbos import WindBos
 from reV.utilities.exceptions import (OutputWarning, ExecutionError,
-                                      OffshoreWindInputWarning)
+                                      OffshoreWindInputWarning,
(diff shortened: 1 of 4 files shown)


reV/econ/econ.py
```diff
@@ -153,6 +159,228 @@ def __init__(self, points_control, cf_file, cf_year, site_data=None,
...
+            inputs in this dataframe, which was split using points_control,
+            only the data relevant to the current sites is passed.
+        econ_fun : method
+            reV_run() method from one of the econ modules (SingleOwner,
+            SAM_LCOE, WindBos).
+        output_request : str | list | tuple
+            Economic output variable(s) requested from SAM.
+        kwargs : dict
+            Additional input parameters for the SAM run module.
+        """
```
grantbuster, line 355: add returns in docstring?

reV/econ/econ.py
```diff
@@ -153,6 +159,228 @@ def __init__(self, points_control, cf_file, cf_year, site_data=None,
...
+
+        if append and cf_file is not None:
+            try:
+                pc = Econ._econ_append_pc(pc.project_points, cf_file,
+                                          sites_per_worker=sites_per_worker)
+            except Exception:
+                msg = ("Failed to adjust ProjectControl for Econ in append "
+                       "mode!")
+                logger.exception(msg)
+                raise
```
grantbuster, line 332: Shouldn't you raise the exception? Otherwise the stderr will just say no active exception to raise?

```
except Exception as e:
    ...
    raise e

File: reV/econ/econ.py [hunk 3]
```diff
@@ -153,6 +159,228 @@ def __init__(self, points_control, cf_file, cf_year, site_data=None,
...
+                        pc = PointsControl.split(
+                            i0, i1, pp, sites_per_split=sites_per_worker)
+                    else:
+                        # this is the scenario where the project points
+                        # is a subset of the meta data like for
+                        # test problems.
+                        pc = PointsControl(
+                            pp, sites_per_split=sites_per_worker)
+
+        except FileNotFoundError as ex:
```
grantbuster, line 274: This is a huge try block for a filenotfound error. Is this because of the complexities of the HSDS and multifile resource handler? Since we only use the resource handler to retrieve the meta, maybe we could just make the try block for just retrieving the meta data, and move the PointsControl methods down and outside of the try block?

reV/econ/econ.py
```diff
@@ -153,6 +159,228 @@ def __init__(self, points_control, cf_file, cf_year, site_data=None,
...
+                        # full generation project points.
+                        i0 = pp.index(gid0)
+                        i1 = pp.index(gid1) + 1
+                        pc = PointsControl.split(
+                            i0, i1, pp, sites_per_split=sites_per_worker)
+                    else:
+                        # this is the scenario where the project points
+                        # is a subset of the meta data like for
+                        # test problems.
+                        pc = PointsControl(
```
grantbuster, line 271: The project points input to this method is extracted from the points control object. Why not just require that points control is the input, then in this case just pass through the PC un-manipulated? 

Also, would this case ever happen? Append mode but the CF gid's are not found in the project points CSV? Should we just raise an exception here? This could be indicative of a mismatch of the project points CSV file and the CF file.

grantbuster (changes requested):

reV/econ/econ.py
```diff
@@ -241,46 +240,27 @@ def _econ_append_pc(pp, cf_file, sites_per_worker=None):
...
-                        # is a subset of the meta data like for
-                        # test problems.
-                        pc = PointsControl(
-                            pp, sites_per_split=sites_per_worker)
-
-        except FileNotFoundError as ex:
-            msg = ('{} is invalid PointsControl can not be adjusted!\n{}'
-                   .format(cf_file, ex))
-            raise ProjectControlFileError(msg)
+        multi_h5_res, hsds = check_res_file(cf_file)
```
grantbuster, line 243: This looks all good. Only suggestion would be to update the doc string to describe what an append-mode points control actually needs to be (just match the single-node generation output meta data).

grantbuster (approved):

PR #226, opened 2020-10-22
Gb/more pv outputs

grantbuster:


reV/SAM/generation.py
```diff
@@ -104,6 +104,12 @@ def _get_res_mean(resource, site, output_request):
                 res_mean = {}
                 res_mean['dni_mean'] = resource['mean_dni', idx] / 1000 * 24
 
+            if 'dhi_mean' in out_req_nomeans:
+                out_req_nomeans.remove('dhi_mean')
+                if res_mean is None:
+                    res_mean = {}
+                res_mean['dhi_mean'] = resource['mean_dhi', idx] / 1000 * 24
```
MRossol, line 111: line ;-)

MRossol (changes requested):
One comment, I think we can clean up the computation of the irradiance_means to be more pythonic

reV/SAM/generation.py
MRossol: There is a lot of redundancy across these if statements to compute the irradiance means

grantbuster: I really disagree with a new line here. We're trying to group the logical blocks. New line after if statement would break up the logical blocks and make it more difficult to read. 

A bit of redundancy but its maybe three or four lines and most of the stuff needs to be hard coded anyway because of the key switch.

MRossol: I was thinking something like this:
```
for irr in ['dni', 'ghi', 'dhi']:
    irr_req = "{}_mean".format(irr)
    if irr_req in out_req_onmeans:
        out_req_nomeans.remove(irr_req)
        if res_mean is None:
            res_mean = {}

        dset = "mean_{}".format(irr)
        res_mean[irr_req] = resource[dset, idx] / 1000 * 24
```

grantbuster: okay! I'm into it.

MRossol (approved):
I approve, merge with admin rights when your ready

PR #229, opened 2020-10-29
Removed input pass through methods in favor of handling the requested

grantbuster:
input pass throughs in the SAM collect outputs method (way more simple
and robust).


MRossol (approved):
Damn impressive!

grantbuster:
I think the impressive part was how convoluted I managed to make this!

Issue #231, opened 2020-11-05
Change stringified dict to actual dict when creating batch scenarios based on CSV input


nickwg03:
**Why this feature is necessary:**
Being able to vary the individual layer arguments within the `excl_dict` would make doing scenario'd runs much easier when we build numerous scenarios that only change arguments within the `excl_dict`. Right now, the process creates a stringified dictionary instead of an actual dictionary. Forcing the dictionary would make the resulting batch JSON usable.

**A possible solution is:**
Force stringified dict into an actual dict

**Additional context**
Below are screenshots of the input batch CSV config, and the batch JSON config with stringified dict.

![df372a1e-c099-4d34-9899-e9e68399f86c](https://user-images.githubusercontent.com/4528512/98294213-cc553f80-1f6c-11eb-9513-2e7bbad015b1.jpg)
![imgo](https://user-images.githubusercontent.com/4528512/98294225-d24b2080-1f6c-11eb-8ab0-2fc4ca4ab276.jpg)

**Charge code**
WFED.10406.19.01.01

**Urgency / Timeframe**
Would need ready to go by Fri, Nov 20th. Earlier is always appreciated :)

MRossol:
implemented in #232

PR #232, opened 2020-11-06
Gb/batch dict

grantbuster:


grantbuster:
implements #231

conda.recipe/meta.yaml
```diff
@@ -16,7 +16,7 @@ requirements:
     - pip
     - setuptools
     - click
-    - h5py
+    - h5py>=3.1.0
```
MRossol, line 19: I tried this in rex and it failed to build, turns out Conda is still at 2.1

reV/batch/batch.py
```diff
@@ -295,15 +295,41 @@ def sub_dirs(self):
...
+            Value to be written to a batch job json file
+
+        Returns
+        -------
+        arg : str | int | float | dict
+            Cleaned value ready to be written to a batch job json file.
+        """
+
+        if isinstance(arg, str):
+            if '{' in arg and '}' in arg:
```
MRossol, line 317: What does the {/} have to do with replacing ' for "?

MRossol (changes requested):
One question, and we need to remove h5py version from the conda.recipe/meta.yaml

Also, I have a bug_fix PR relation to a rex release. We should coordinate merging these two PRs...

conda.recipe/meta.yaml
grantbuster: ah i see. Fixed

reV/batch/batch.py
grantbuster: the braces look for a dictionary that has been stringified. The ' vs " replacement has to do with how json wants strings with the double quotes.

grantbuster:
@MRossol removed the h5py version from conda. Let me know how you want to merge the two PR's. I incremented the rev version on this branch. Both should be easy to rebase.

MRossol:
Lets merge this one and then I can rebase mine while the tests run

Issue #230, opened 2020-10-30
Forced Inclusion Layer


MRossol:
**Why this feature is necessary:**
Add a “forced_inclusion” layer functionality in reV. Essentially, apply all exclusions except in certain locations given an input layer.

**A possible solution is:**
Add new LayerMask type called forced_inclusion_values to https://github.com/NREL/reV/blob/13919192da0478375a06246afeb0643caad299ba/reV/supply_curve/exclusions.py#L16

Apply forced_inclusion_values layers at the end of: https://github.com/NREL/reV/blob/13919192da0478375a06246afeb0643caad299ba/reV/supply_curve/exclusions.py#L739-L747

**Additional context**
One thing that has tripped up ReEDS is when we exclude too much resource within a given region based on our siting regimes. In addition, when we do any back mapping (ReEDS->reV), ReEDS has to guess where existing wind farms are – actually does a decent job of this but it’s not perfect.

**Charge code**
WEPG.10348.08.01.01

**Urgency / Timeframe**
FY21 Q1

MRossol:
Implemented in #234

PR #234, opened 2020-11-09
add force_inclusion_values

MRossol:
PR for #230 
- Added force_inclusion_values to Layers class
- Added a new _force_include attr and property
- Isolate force_include layers and run them after regular layers, use np.maximum instead of np.minimum to force include layer values.

### DIFF
--- a/reV/supply_curve/exclusions.py
+++ b/reV/supply_curve/exclusions.py
@@ -789,7 +789,7 @@ class ExclusionMask:
                     else:
                         mask = np.minimum(mask, layer_mask)
 
-            mask = self._force_include(mask, force_include)
+            mask = self._force_include(mask, force_include, ds_slice)
 
             if self._min_area is not None:
                 mask = self._area_filter(mask, min_area=self._min_area,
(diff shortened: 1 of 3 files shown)


reV/supply_curve/exclusions.py
```diff
@@ -710,6 +730,27 @@ def _generate_ones_mask(self, ds_slice):
...
+    def _force_include(self, mask, layers, ds_slice):
+        """
+        Apply force inclusion layers
+
+        Parameters
+        ----------
+        mask : ndarray | None
+            Mask to apply force inclusion layers to
+        layers : list
+            List of force inclusion layers
```
grantbuster, line 742: add ds_slice docstring?

grantbuster (changes requested):
Nice this looks great! One minor doc string comment and can you also add the force_include option to the SC agg CLI help string under the excl_dict input?

reV/supply_curve/exclusions.py
MRossol: whoops

MRossol:
Doc strings updated

grantbuster (approved):

PR #236, opened 2020-11-12
Gb/gen site data

grantbuster:


reV/SAM/SAM.py
```diff
@@ -629,6 +636,24 @@ def get_time_interval(time_index):
...
+        Parameters
+        ----------
+        site_sys_inputs : dict
+            Optional set of site-specific SAM system inputs to complement the
+            site-agnostic inputs.
+        """
+
+        if site_sys_inputs is not None:
+            for k, v in site_sys_inputs.items():
+                if isinstance(v, float):
```
MRossol, line 651: You should be able to wrap all of this logic into a single if statement
Also what happens if the value is None, we just don't set it? Area all sam_sys_inputs defaulted to None?

MRossol (changes requested):
Looks good. Like the changes. One question and comment

reV/SAM/SAM.py
grantbuster: can you suggest the if statement? It's tricky because `if isinstance(v, float) and ~np.isnan(v):` doesn't work because then a NaN values gets sent to the `else:` statement. 

None values will get set although I've never seen these used. Because the site_data gets read in as a dataframe the missing values get filled with NaN.

MRossol: Two options in order of my preference:
1) 
```
        if site_sys_inputs is not None:
            for k, v in site_sys_inputs.items():
                if isinstance(v, float) and np.isnan(v):
                    v = None
                
                self.sam_sys_inputs[k] = v
```
2) 
```
        if site_sys_inputs is not None:
            for k, v in site_sys_inputs.items():
                if isinstance(v, float) and np.isnan(v):
                    continue
                else:
                    self.sam_sys_inputs[k] = v
```

grantbuster: Okay i went with 2 because in 1 if v is NaN we don't want to overwrite whatever is already in `sam_sys_inputs`

reV/SAM/SAM.py
```diff
@@ -648,9 +648,8 @@ def _parse_site_sys_inputs(self, site_sys_inputs):
 
         if site_sys_inputs is not None:
             for k, v in site_sys_inputs.items():
-                if isinstance(v, float):
-                    if ~np.isnan(v):
-                        self.sam_sys_inputs[k] = v
+                if isinstance(v, float) and ~np.isnan(v):
```
MRossol, line 651: I think this needs to be np.isnan(v) and continue not pass

MRossol (approved):

grantbuster:
implements #235

Issue #240, opened 2020-11-17
Add clipping outputs


nickwg03:
**Why this feature is necessary:**
For the PV+battery modeling, we’d like to know the power being clipped by the inverter, as that power can be sent to the battery without affecting the PV-to-grid output. I thought we’d be able to calculate the clipped power in hourlize post-processing, but since the reV outputs are aggregated over multiple sites (some of which may be clipped in a given timestamp and others not), I can’t extract a definitive clipping estimate from the aggregated profiles. If I try using the same method for the aggregated profiles as for a single pixel I end up with ~50% lower clipping.

Could we add clipped power as an additional profile output from reV?

**A possible solution is:**
One liner:
`clipped = np.where(ac < ac.max(), 0, dc - ac)`

**Charge code**
GMLC.10246.09.05.01

**Urgency / Timeframe**
Not urgent, but within the next month would be helpful

MRossol:
@dguittet does SAM already calculate clipping on a timestep by timestep basis? If not is there any easy way to get clipping out of SAM or should we just compute it in reV as Grue requested?

dguittet:
It doesn't for PVWattsv7 but does for Pvsamv1. I don't think the solution above would work because the PVwatts `ac` output is post-clipping.

I can add a clipping time series output to PVWattsv7 for our imminent release.

dguittet:
Also happy to support and review a PR if someone wants to add time series inverter losses to the PVWatts outputs :)

grantbuster:
@dguittet how would we do that? Is there a clipping limit input to pvwatts? I was thinking we would have to run a simulation for both the model with and without clipping then diff the profiles. This would be difficult in our architecture so thats why we're reaching out to see if you have access to the "diff" on the backend.

patrickbrown4:
Since the `ac` output is post-clipping, shouldn't `dc - ac` give the power that is forgone due to clipping? The `clipped` profile calculated using `clipped = np.where(ac < ac.max(), 0, dc - ac)` looks like what I'd expect when applied for a single NSRDB pixel: https://www.evernote.com/l/AKpG4LiF_WpJ57R74ygivcCe9ATtWH--6Vo

grantbuster:
[...]

PR #243, opened 2020-11-25
Clipping

MRossol:
Implementation of #240


grantbuster:
@MRossol  can you also add the new variable here: https://github.com/NREL/reV/blob/b7fc71446f1159a203372532f2acdeb0ecc8a123/reV/config/output_request.py#L62

MRossol:
I knew I missed something!

reV/SAM/generation.py
```diff
@@ -589,22 +589,39 @@ def ac(self):
...
+        Returns
+        -------
+        clipped : np.ndarray
+            1D array of clipped DC power in kW.
+            Datatype is float32 and array length is 8760*time_interval.
+        """
+        ac = self.ac()
+        dc = self.dc()
+
+        return np.where(ac < ac.max(), 0, dc - ac)
```
grantbuster, line 623: Why wouldnt we always want `dc - ac`? Wouldnt we want the clipped power at hour 11 in the image below? 

image

patrickbrown4: On the ReEDS side, we want to differentiate between 1) sub-clipping power that can go either to the grid through the inverter or to storage, and 2) clipped power that can only go to storage. Based on the Texas profiles I was looking at (https://www.evernote.com/l/AKpG4LiF_WpJ57R74ygivcCe9ATtWH--6Vo), when multiple NSRDB pixels are averaged you get some pixels that are clipping in a certain hour and some that aren't, so you can't pull out the clipped power. For hour 11, the ratio of `dc`/`ac` power is just 1/efficiency_inverter, which we can model on our side from the `dc` profile for power that goes to the grid.

I realize now that if we do representative rather than average profiles for each reV site, the "some-pixels-are-clipping-while-others-aren't" issue would go away and we could get the clipped profile in hourlize post-processing. But it would still be nice to generate it directly from reV in case someone wants to use average profiles.

grantbuster: So we always want to represent clipped power as `np.where(ac < ac.max(), 0, dc - ac)` then? Or is there another term for just the plain `dc - ac` clipping that we should add?

grantbuster (approved):

reV/SAM/generation.py
patrickbrown4: Since we already have `dc` and `ac` outputs, we can get `unclipped_dc` from `dc - clipped_power`. I checked and the profiles are the same whether this step is performed before or after the aggregation.

PR #244, opened 2020-12-01
Gb/downscaling

grantbuster:


MRossol (approved):
why change the internal variable to ds_list? Seems ambiguous compared to downscale?

reV/config/sam_config.py
```diff
@@ -127,25 +127,33 @@ def downscale(self):
...
         """
         if self._downscale is None:
-            downscale = []
+            ds_list = []
             for v in self.inputs.values():
-                downscale.append(v.get('downscale', None))
+                ds_list.append(v.get('downscale', None))
 
-            self._downscale = list(set(downscale))
+            self._downscale = ds_list[0]
```
grantbuster, line 144: @MRossol, ds_list is just a temporary variable that gets the list of all the downscaling arguments for all SAM input json files and makes sure they're the same. self._downscale is the attribute that gets used outside of the property and is set here.

PR #246, opened 2020-12-08
Gb/pipeline mods

grantbuster:


MRossol (approved):
Looks good, NOTE: I never released 0.4.37 so we can stay on that version

grantbuster:
oh sweet! Dropped that commit.

MRossol (approved):

PR #249, opened 2020-12-16
bug fix on offshore transmission table input with overlapping column …

grantbuster:
…labels vs the sc points table


MRossol (approved):
Looks good, shame we have to deal with this when this all is going to get blown up... Just not soon enough

PR #252, opened 2021-01-05
add mean_wind_dirs method to points

MRossol:
@grantbuster, thoughts on a test for this? Unfortunately, the ri_exclusions.h5 is mapped to the generation file. I could create a new file with a TM dataset mapped to the resource gids...


MRossol:
@grantbuster  nevermind, I can just add the gen gid to the meta

reV/supply_curve/points.py
```diff
@@ -466,6 +467,37 @@ def exclusion_weighted_mean(self, arr, drop_nan=True):
...
+
+        else:
+            arr_slice = gids
+            ax = 0
+
+        angle = np.radians(arr[arr_slice], dtype=np.float32)
+        sin = np.mean(np.sin(angle) * excl, axis=ax)
+        cos = np.mean(np.cos(angle) * excl, axis=ax)
+
+        mean_wind_dirs = np.degrees(np.arctan2(sin, cos))
```
grantbuster, line 497: only comment is that it looks like this set of functions will give us degrees in range -180 to 180. Is this what we want?

MRossol: Great catch! No we want 0-359. Let me check the WTK code as I did this there...

MRossol: @grantbuster this should be fixed, my checking my math?

grantbuster: Yeah looks good!

grantbuster (approved):

PR #253, opened 2021-01-06
Economy of scale

grantbuster:


reV/SAM/generation.py
```diff
@@ -228,41 +228,42 @@ def collect_outputs(self, output_lookup=None):
...
-                raise KeyError('Cannot request more than one single owner '
-                               'output in Generation module. Found the '
-                               'following {} single owner output requests in '
-                               'the generation run: {}'
-                               .format(len(reqs), reqs))
-            elif len(reqs) == 1:
-                so_out_req = self.output_request.pop(
-                    self.output_request.index(reqs[0]))
-
+            so_vars = ('ppa_price', 'lcoe_real', 'lcoe_nom',
```
MRossol: Where is single owner turned on?

reV/econ/economies_of_scale.py
```diff
@@ -0,0 +1,306 @@
...
+        """LCOE calculated with the unscaled (raw) capital cost
+
+        LCOE = (FCR * raw_capital_cost + FOC) / AEP + VOC
+
+        Returns
+        -------
+        lcoe : float | np.ndarray
+            LCOE calculated with the unscaled (raw) capital cost
+        """
+        lcoe = ((self.fcr * self.raw_capital_cost + self.foc) / self.aep
```
MRossol, line 285: Do we want to make LCOE a static method with the required args and call it here and in the below property?

reV/generation/base.py
```diff
@@ -34,6 +34,71 @@ class BaseGen(ABC):
     # Mapping of reV generation / econ outputs to scale factors and units.
     OUT_ATTRS = {}
 
+    # Mapping of reV econ outputs to scale factors and units.
+    # Type is scalar or array and corresponds to the SAM single-site output
+    # This is the OUT_ATTRS class attr for Econ but should also be accessible
+    # to rev generation
+    ECON_ATTRS = {'other': {'scale_factor': 1, 'units': 'unknown',
```
MRossol, line 41: Not essential for this PR, but I wonder if we should think about moving teh ECON and GEN attrs to JSONs and then adding in kwargs. If the kwargs are None the defaults are imported. I'm just wondering if anyone might want to change the attrs...

reV/handlers/cli_multi_year.py
```diff
@@ -80,7 +80,7 @@ def from_config(ctx, config_file, verbose):
 
     # initialize loggers.
     init_mult(name, config.logdir,
-              modules=[__name__, 'reV.handlers.multi_year'],
+              modules=[__name__, 'reV'],
```
MRossol, line 83: I forgot about the handler CLIs, good catch!

MRossol (commented):
The structure looks good, a few style comments

reV/SAM/generation.py
grantbuster: The block of code you commented on is where generation has the possibility for running econ as a sort of tag-along computation. The first block of code checks for SO output requests and the second block actually runs the SO as a tag-along:

https://github.com/NREL/reV/blob/c30a64955934cc4fe3d9e44beb349b9ce43c1165/reV/SAM/generation.py#L242-L245

https://github.com/NREL/reV/blob/c30a64955934cc4fe3d9e44beb349b9ce43c1165/reV/SAM/generation.py#L264-L271

reV/econ/economies_of_scale.py
grantbuster: Yeah that's not a bad idea in case someone just wants the LCOE equation. Where do you think is a good place for the static method? I almost think its most appropriate in the Econ class even though it would only get used here.

reV/generation/base.py
grantbuster: Yeah i agree... Remind me, if we put json files in sub directories in the reV module directories will they get copied over when pip installed?

reV/econ/economies_of_scale.py
MRossol: ugg, thats a good question. My first thought was in the EconomyOfScale class, but now that you mention it what about putting it in utilities? If we ever envision it being a stand-alone function, might as well make it one.

reV/handlers/cli_multi_year.py
grantbuster: I only fixed this here, Let me also fix it in the collection cli...

reV/generation/base.py
MRossol: yup, just like the SAM default .csvs, we just have to add to this list:
https://github.com/NREL/reV/blob/ce5ef74aabd6398895e03017143d6aa560fd013c/setup.py#L78

grantbuster: Done, can you check this commit? I figure we can add something to modify the attributes on the fly later. 

https://github.com/NREL/reV/pull/253/commits/9c3e4f9db0156f8f4a60a0f40d94d3873ff0d99e

grantbuster: Now this commit (after rebase): https://github.com/NREL/reV/pull/253/commits/bf05da6c1fc06b444af97bad2b6481d5e61a7e69

MRossol (approved):
Looks great!

PR #256, opened 2021-01-13
Excl scale factor

MRossol:
Use Resource handler instead of h5py in ExclusionLayers to allow for use of scale_factors in exclusion layers

### DIFF
--- a/.github/workflows/conda_build.yml
+++ b/.github/workflows/conda_build.yml
@@ -19,7 +19,6 @@ jobs:
       env:
         ANACONDA_TOKEN: ${{ secrets.ANACONDA_TOKEN }}
       run: |
-        conda install --yes --quiet conda-build conda-verify
-        conda install --yes --quiet anaconda-client
+        conda install --yes --quiet conda-build conda-verify anaconda-client
         conda build conda.recipe/ --channel=nrel
         anaconda -t $ANACONDA_TOKEN upload -u nrel $(conda build conda.recipe/ --channel=nrel --output)
--- a/conda.recipe/meta.yaml
+++ b/conda.recipe/meta.yaml
@@ -15,7 +15,7 @@ requirements:
     - python
     - pip
     - setuptools
-    - NREL-rex>=0.2.38
+    - NREL-rex>=0.2.42
     - NREL-PySAM==2.1.4
     - packaging
     - plotting
@@ -23,7 +23,7 @@ requirements:
   run:
     - python
     - pip
-    - NREL-rex>=0.2.38
+    - NREL-rex>=0.2.42
     - NREL-PySAM==2.1.4
     - packaging
     - plotting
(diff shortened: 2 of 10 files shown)


grantbuster (approved):

MRossol:
@grantbuster,
I'm getting a bug in your econ_of_scale tests: https://github.com/NREL/rex/runs/1698009185?check_suite_focus=true#step:7:395

I'm a bit confused as to whats causing the two different dtypes. when we pull the exclusion layers they should get converted to a 'float32': https://github.com/NREL/reV/blob/7e49d3e4c7a3aabb6e8d2fddc052b68f809c9f36/reV/supply_curve/exclusions.py#L216

this is the meat:
``
self = <reV.supply_curve.point_summary.SupplyCurvePointSummary object at 0x7f2611e13670>
flat_arr = array([53455000, 53455000, 53455000, 53455000, 53455000, 53455000,
       53455000, 53455000, 53455000, 53455000, 5345...000, 53455000, 53455000,
       53455000, 53455000, 53455000, 53455000, 53455000, 53455000,
       53455000, 53455000])

    def exclusion_weighted_mean(self, flat_arr):
        """Calc the exclusions-weighted mean value of a flat array of gen data.
    
        Parameters
        ----------
        flat_arr : np.ndarray
            Flattened array of resource/generation/econ data. Must be
            index-able with the self._gen_gids array (must be a 1D array with
            an entry for every site in the generation extent).
    
        Returns
        -------
        mean : float
            Mean of flat_arr masked by the binary exclusions then weighted by
            the non-zero exclusions.
        """
        x = flat_arr[self._gen_gids[self.bool_mask]]
>       x *= self.excl_data_flat[self.bool_mask]
E       numpy.core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'multiply' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
```

Mind taking a look?

FYI I fixed the other failed tests will push to this PR now

grantbuster:
@MRossol the `ResourceDataset._unscale_data()` in rex converts the data to float32. Since exclusion files dont have scale_factor attributes theyre no longer being converted to float32.

grantbuster:
I suggest converting the exclusion data to float32 in the appropriate reV exclusion class.

MRossol:
That is where I'm confused. I thought we did that here:
https://github.com/NREL/reV/blob/7e49d3e4c7a3aabb6e8d2fddc052b68f809c9f36/reV/supply_curve/exclusions.py#L216

grantbuster:
Did the conda-build fix work? Any idea why reVX is failing now?

MRossol:
@grantbuster, I haven't tried the conda_build fix yet, I'll try it with reV. I did get rex to build locally which was my work around.
reVX is failing because of a rounding error issue. I've fixed it on the reVX branch. Okay to merge this now that reV is passing?

grantbuster (approved):

grantbuster:
yep go for it!

PR #262, opened 2021-01-27
multiply data by excl_multi prior to data_layer aggregation

### LINKED ISSUE #261: Supply-Curve-Aggreation: Take partial exclusions into account for data_layer characterization sums
**Bug Description**
In supply-curve-aggregation, the "sum" characterization method for data_layers appears to be adding the pixel counts within a supply curve point before removing partial exclusions. We would like this figure to allow us to calculate the area of non-excluded land within each characterization.

**Expected behavior**
The sum of a characterization layer is a float taking into account partial exclusions.

**System (please complete the following information):**
 - OS: Linux
 - Compute hardware Eagle
 - Version: v0.4.43

MRossol:
closes #261


reV/supply_curve/point_summary.py
```diff
@@ -428,6 +379,153 @@ def mean_h5_dsets_data(self):
...
+                     'match excl mult!')
+                logger.error(e)
+                raise DataShapeError(e)
+
+            if method == 'category':
+                data = method_func['category'](data, excl_mult)
+            elif method == 'mode':
+                data = method_func['mode'](data)
+            else:
+                data = data * excl_mult
```
grantbuster, line 475: Did you see our thread on teams? I dont think (max, min, mode) should be weighted. I do think that (mean, sum, categorize) should be weighted.

reV/supply_curve/point_summary.py
```diff
@@ -428,6 +379,153 @@ def mean_h5_dsets_data(self):
 
         return _mean_h5_dsets_data
 
+    @staticmethod
+    def _mode(data):
```
grantbuster, line 383: This seems super unnecessary

reV/supply_curve/point_summary.py
```diff
@@ -113,55 +113,6 @@ def __init__(self, gid, excl, gen, tm_dset, gen_index, excl_dict=None,
 
         self._apply_exclusions()
 
-    def _apply_exclusions(self):
```
grantbuster, line 116: Just to confirm, this method and _resource_exclusion() were not modified, just moved?

grantbuster (changes requested):

reV/supply_curve/point_summary.py
MRossol: Missed that, will update

reV/supply_curve/point_summary.py
MRossol: I agree its a bit silly, but I had to wrap it to enable this dictionary:
https://github.com/NREL/reV/blob/94e9bab80dc3e46d8d1620a9b9199c2e9e6eac0d/reV/supply_curve/point_summary.py#L446-L451

reV/supply_curve/point_summary.py
MRossol: yeah just re-organized a bit

grantbuster (approved):

PR #280, opened 2021-03-08
Exclusions refactor

MRossol:
add contains to exclusions handler
simplify techmap args and kwargs


grantbuster (dismissed):
looks good if the tests pass!

PR #281, opened 2021-03-10
refactor tech_mapping

MRossol:
@grantbuster I had to update the baseline techmap due to removing the ability to set the distance threshold. Now it's computed with a user-defined dist_margin (default 1.05).
Plotted below is the difference between the old baseline and the new techmap (black dots).  The new techmap is slightly more conservative (i.e., more -1 pixels)

Figure_1

### DIFF
--- a/reV/supply_curve/points.py
+++ b/reV/supply_curve/points.py
@@ -1469,6 +1469,67 @@ class SupplyCurveExtent:
 
         return self._points
 
+    @staticmethod
+    def _chunk_excl(arr, resolution):
+        """Split an array into a list of arrays with len == resolution.
+
+        Parameters
+        ----------
+        arr : np.ndarray
+            1D array to be split into chunks.
+        resolution : int
+            Resolution of the chunks.
+
+        Returns
+        -------
+        chunks : list
+            List of arrays, each with length equal to self.resolution
+            (except for the last array in the list which is the remainder).
+        """
+
+        chunks = []
+        i = 0
+        while True:
+            if i == len(arr):
+                break
+            else:
+                chunks.append(arr[i:i + resolution])
+
+            i = np.min((len(arr), i + resolution))
+
+        return chunks
+
+    @staticmethod
+    def _excl_slices(arr, resolution):
+        """Split row or col ind into slices of excl rows or slices
+
+        Parameters
+        ----------
+        arr : np.ndarray
+            1D array to be split into slices
+        resolution : int
+            Resolution of the sc points
+
+        Returns
+        -------
+        slices : list
+            List of arr slices, each with length equal to self.resolution
+            (except for the last array in the list which is the remainder).
+        """
+
+        slices = []
+        i = 0
+        while True:
+            if i == len(arr):
+                break
+            else:
+                s, e = arr[i:i + resolution][[0, -1]]
+                slices.append(slice(s, e + 1))
+
+            i = np.min((len(arr), i + resolution))
+
+        return slices
+
     def get_excl_slices(self, gid):
         """Get the row and column slices of the exclusions grid corresponding
         to the supply curve point gid.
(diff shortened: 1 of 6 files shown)


grantbuster:
The map looks great! a bit of difference on the edges of the resource domain is fine. I think we should plot up the full techmap at some point to make sure there's no gaps though. The test case is pretty ideal without terrible skew to the resource grid.

reV/supply_curve/points.py
```diff
@@ -1469,6 +1469,67 @@ def points(self):
...
+
+        chunks = []
+        i = 0
+        while True:
+            if i == len(arr):
+                break
+            else:
+                chunks.append(arr[i:i + resolution])
+
+            i = np.min((len(arr), i + resolution))
```
grantbuster, line 1498: can we do this with one of the numpy array split methods instead of a while loop with append?

Sorry i know i might have written this code but since you moved it it looks "changed" and therefore i get to make annoying refactor suggestions :)

reV/supply_curve/points.py
```diff
@@ -1469,6 +1469,67 @@ def points(self):
...
+        slices = []
+        i = 0
+        while True:
+            if i == len(arr):
+                break
+            else:
+                s, e = arr[i:i + resolution][[0, -1]]
+                slices.append(slice(s, e + 1))
+
+            i = np.min((len(arr), i + resolution))
```
grantbuster, line 1529: Same for this, ive recently been using a numpy array split then a list comprehension where `slices = [slice(chunk[0], 1 + chunk[-1]) for chunk in array_split]`

reV/supply_curve/tech_mapping.py
```diff
@@ -275,138 +232,173 @@ def map_resource_gids(cls, gids, excl_fpath, res_fpath,
...
+        logger.debug('Getting tech map coordinates for chunks {} through {}'
+                     .format(gids[0], gids[-1]))
+        ind_out = []
+        coords_out = cls._get_excl_coords(excl_fpath, gids, points,
+                                          excl_row_slices, excl_col_slices)
 
-        logger.debug('Getting tech layer coordinates for chunks {} through {}'
+        logger.debug('Running tech mapping for chunks {} through {}'
                      .format(gids[0], gids[-1]))
+        for i, _ in enumerate(gids):
```
grantbuster, line 268: much more simple, nice!

grantbuster (approved):
Looks good! A few minor nitpicky suggestions on code that i probably wrote a few years ago haha. The million dollar question is: were you able to run this for your new exclusions/resource? Did you fix the -1 issues? Is the refactored code faster than previous?

MRossol:
So the -1 issue was due to Travis' lat, lons.

Performance on the test domain:
Master:
In [4]: %%time
   ...: ind = TechMapping.run(EXCL, RES, dset=None, max_workers=2)
   ...:
   ...:
CPU times: user 166 ms, sys: 188 ms, total: 354 ms
Wall time: 4.89 s

New branch:
In [2]: %%time
   ...: ind = TechMapping.run(EXCL, RES, dset=None, max_workers=2)
   ...:
   ...:
CPU times: user 78.4 ms, sys: 68.4 ms, total: 147 ms
Wall time: 3.89 s

Not a great test case, but still faster!

grantbuster:
Sweet! if it works, its faster, and its more simple then hell yes.

reV/supply_curve/points.py
MRossol: So I can update this, BUT, the results are slightly different:
```
In [26]: for i in range(len(baseline)):
    ...:     print(i)
    ...:     print(len(baseline[i]))
    ...:     print(len(test[i]))
    ...:     print()
    ...: 
0
64
63

1
64
63

2
64
63

3
64
63

4
64
63

5
64
63

6
64
63

7
64
63

8
64
62

9
64
62

10
64
62

11
64
62

12
64
62

13
64
62

14
64
62

15
40
62
```

MRossol: FYI:
```
test = np.array_split(arr, ceil(len(arr) / resolution))
```

reV/supply_curve/points.py
MRossol: what about this:
```
        slices = get_chunk_ranges(len(arr), resolution)
        slices = list(map(lambda i: slice(*i), slices))
 ```

File: reV/supply_curve/points.py (hunk 1 above)
Comment (MRossol) [reply reV/supply_curve/points.py:1498]: If we want to retain the chunk sizes from the existing function we could do this:
```
        chunks = get_chunk_ranges(len(arr), resolution)
        chunks = list(map(lambda i: np.arange(*i), chunks))
```

PR #286, opened 2021-03-23
Sam config dict

MRossol:
Allow passing of SAM config as a dict

### DIFF
--- a/reV/config/project_points.py
+++ b/reV/config/project_points.py
@@ -688,6 +688,10 @@ class ProjectPoints:
             logger.error(msg)
             raise ConfigError(msg)
 
+        if len(df_configs) == 1 and df_configs[0] is None:
+            self._df['config'] = list(sam_configs)[0]
+            df_configs = self.df['config'].unique()
+
         # Check to see if config references in project_points DataFrame
         # are valid file paths, if compare with SAM configs
         # and update as needed
--- a/reV/config/project_points.py
+++ b/reV/config/project_points.py
@@ -212,7 +212,7 @@ class ProjectPoints:
     >>> h_list = pp.h
     """
 
-    def __init__(self, points, sam_config, tech=None, res_file=None,
+    def __init__(self, points, sam_configs, tech=None, res_file=None,
                  curtailment=None):
         """
         Parameters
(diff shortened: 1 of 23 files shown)


MRossol:
@grantbuster, Okay this should be good to go, lots of test issues

reV/config/project_points.py
```diff
@@ -676,6 +448,166 @@ def h(self):
...
+            raise TypeError('Project Points sites needs to be set as a list, '
+                            'tuple, or slice, but was set as: {}'
+                            .format(type(points)))
+
+        df['config'] = None
+
+        return df
+
+    @classmethod
+    def _parse_points(cls, points, res_file=None):
```
grantbuster, line 534: Whats the motivation for moving all these methods around? I'm all for re-org but it makes reviewing the diff's impractical.

tests/test_config.py
```diff
@@ -241,6 +242,23 @@ def test_duplicate_coords():
         ProjectPoints.regions(regions, res_file, sam_files)
 
 
+def test_sam_configs():
```
grantbuster, line 245: If you have time, maybe consider adding a negative tests of the bad sam_files input types? Like a list/tuple of sam filepaths and a single extracted SAM config dictionary (instead of an id-to-config mapper).

grantbuster (changes requested):

reV/config/project_points.py
MRossol: Yeah sorry on the diff issues, It was re-org for readability

tests/test_config.py
MRossol: will do, looks like there are some other tests failing... I'll dig into that.

MRossol: @grantbuster, miss any big ones to test for?
https://github.com/NREL/reV/blob/2de9e54ecd6b2d6196a8223647f37d5ec4ce2e40/tests/test_config.py#L262

grantbuster: I'm not sure, can you annotate what youre testing for in these? I dont see why the 1st, 2nd, and 3rd tests should raise an error? And yes can you add the case where there is a list of sam_files input? I think that was one of our previous use cases that we deprecated right?

grantbuster: Ah I see. @MRossol looks good!

grantbuster (approved):

PR #287, opened 2021-03-29
Gb/multi excl file

grantbuster:


reV/config/supply_curve_configs.py
```diff
@@ -43,8 +43,15 @@ def __init__(self, config):
...
-        with h5py.File(self.excl_fpath, mode='r') as f:
-            dsets = list(f)
+        dsets = []
+        paths = self.excl_fpath
+
+        if isinstance(self.excl_fpath, str):
+            paths = [self.excl_fpath]
+
+        for fp in paths:
+            with h5py.File(fp, mode='r') as f:
```
MRossol, line 53: I'm not saying that this isn't nice and clean, but did you think about just doing:
```
with MultiFileResource(paths) as f:
    dsets = f.datasets
```

I'm honestly not sure if it is cleaner on the back end or not

MRossol (dismissed):

reV/config/supply_curve_configs.py
grantbuster: Yeah that works! MultiFileResource can handle a list of just one file right? That's the only reason why i used a single File object but I forgot that we could just use MultiFileResource with a single member list

MRossol: I would think so but I haven't tested it...

MRossol (approved):

Issue #241, opened 2020-11-19
Integrate ORCA into reV Econ


MRossol:
**Why this feature is necessary:**
- Detangle ORCA from reV. 
- Directly import cost equations from new Repo.
- Build offshore CAPEX and OPEX calculator (analog to WindBos)
- Push CAPEX and OPEX to SAM

**Charge code**
WEPG 10348 08.01.03

**Urgency / Timeframe**
Q1

PR #295, opened 2021-04-20
Nrwal integration

grantbuster:


reV/config/offshore_config.py
```diff
@@ -51,6 +51,23 @@ def sam_files(self):
         """Get the sam files dict"""
         return self['sam_files']
 
+    @property
+    def nrwal_configs(self):
+        """Get the nrwal configs dict"""
+        return self['nrwal_configs']
+
+    @property
+    def offshore_meta_cols(self):
```
MRossol, line 60: Is this now going to be values from the offshore/NRWAL inputs?  Maybe re-name it NRWAL_input_cols?

reV/config/pipeline.py
```diff
@@ -86,18 +86,28 @@ def _check_dirout_status(self):
...
-        _pipeline_steps : list
+        pipeline : list
             reV pipeline run steps. Should be a list of (command, config)
             pairs.
         """
 
         return self['pipeline']
 
+    @property
+    def logging(self):
```
MRossol, line 102: Should we move this to the base config class? Seems useful!

reV/offshore/offshore.py
```diff
@@ -165,75 +169,172 @@ def _parse_cf_meta(gen_fpath):
...
+            if 'turbine_capacity' not in sys_in:
+                msg = ('System input key "turbine_capacity" not found in '
+                       'SAM system inputs for "{}". Calculating from turbine '
+                       'power curves.'.format(cid))
+                logger.warning(msg)
+                warn(msg, OffshoreWindInputWarning)
+
+            loss1 = sys_in.get('wind_farm_losses_percent', 0)
+            loss2 = sys_in.get('turb_generic_loss', 0)
+            if loss1 != 0 or loss2 != 0:
```
MRossol, line 283: since the loses are a scalar hair-cut why not raise the warning and remove the losses that were applied?

reV/offshore/offshore.py
```diff
@@ -262,597 +363,200 @@ def meta_source_offshore(self):
...
+                logger.error(msg)
+                logger.error('This is the offshore meta that is causing NaN '
+                             'outputs: {}'.format(nan_meta))
+                logger.error('These are the resource gids causing NaN '
+                             'outputs: {}'.format(nan_gids))
+                raise ValueError(msg)
+
+    def write_to_gen_fpath(self):
+        """Save offshore outputs to input generation fpath file. This will
+        overwrite data!"""
```
MRossol, line 466: We should add this as a log statement

reV/offshore/orca.py
```diff
@@ -1,145 +0,0 @@
-# -*- coding: utf-8 -*-
```
MRossol, line 1: MUWAHAHAHAHAH

reV/supply_curve/exclusions.py
```diff
@@ -735,7 +735,20 @@ def _generate_ones_mask(self, ds_slice):
...
+        ones_shape = ()
+        for i, s in enumerate(self.shape):
+            try:
+                ax_slice = ds_slice[i]
+                if np.issubdtype(type(ax_slice), np.integer):
+                    ones_shape += (ax_slice,)
+                else:
+                    ax = np.arange(s, dtype=np.int32)
+                    ones_shape += (len(ax[ax_slice]), )
+            except IndexError:
```
MRossol, line 748: I need to test this a bit more, I'm worried that if ax_slice is a list or ndarray with idxs outside the range of ax it will cause and index error... I'll proiritize that tomorrow, mind if I push a commit if needed?

reV/supply_curve/point_summary.py
```diff
@@ -627,6 +622,9 @@ def point_summary(self, args=None):
                 'timezone': self.timezone,
                 }
 
+        if self.offshore is not None:
+            ARGS['offshore'] = self.offshore
```
MRossol, line 626: Why do we need to track offshore still?

MRossol (commented):
A few questions and I want to check that new exclusions one method

reV/config/offshore_config.py
grantbuster: yes its offshore data inputs that get piped to the output meta data. I named it to imply that its meta data for the offshore sites that needs to be included in the output meta data. I think the name is okay since it ties this to the meta dataframe?

reV/config/pipeline.py
grantbuster: Pipeline and batch use these logging dictionaries but everything else uses a combination of log_level at the base level and a logging directory in the directories namespace. I'm not totally opposed but it would require a full rework of all the config organizations and would break backwards compatibility.

reV/offshore/offshore.py
grantbuster: we could do that.... I'm not sure its worth the complexity and the implicit manipulation of the data. There are also a ton of different loss terms now in the sam turbine configs that we would have to check for and combine.... I'm not crazy about the idea.....

reV/offshore/offshore.py
grantbuster: Like this? Will do.

```
logger.info('Saving offshore outputs to reV output file "{}". This will overwrite data!')

reV/supply_curve/exclusions.py
grantbuster: definitely. There's already a test framework here you can work on, that's how I fixed some edge cases: `test_supply_curve_exclusions.py::test_no_excl`

reV/supply_curve/point_summary.py
grantbuster: I figured it is definitely worthwhile to have an offshore flag (0 or 1) in the aggregation output table if nothing else than to just filter the offshore sites easily. We dont treat those points any differently but its more of a meta data tracking thing.

reV/config/offshore_config.py
MRossol: Sounds good to me, maybe just clarify in teh docstring, something along the lines of:
"columns from offshore/NRWAL input data to be passed through to the resource meta data table"

reV/config/pipeline.py
MRossol: gotcha, I think this is fine then even if its in two places.

reV/offshore/offshore.py
MRossol: You sold me at a ton of new loss terms...

reV/supply_curve/point_summary.py
MRossol: Ah gotcha so this is just for the sc/agg output table, agreed there!

reV/config/offshore_config.py
grantbuster: for sure, the docstring for the offshore module is much better, i see now that I slacked on the config and cli docstring! Will edit.

MRossol: LOL, no worries man, I've been more than guilty of that ;-)

reV/supply_curve/exclusions.py
MRossol: FYI, as I suspected there was a fringe case, fixed it with and if/else instead of try/except. Will fail now if the ds_slice causes an index error

MRossol:
closes #241

Issue #290, opened 2021-04-06
Clean up Gen tests


MRossol:
**Bug Description**
@mikebannis  your Generation tests keep failing every time the PySAM version changes. Can you refactor them to be more robust. I'd suggest removing hard-coded "true" values and instead caching baseline results from PySAM 2.1.4 and then use np.allclose or pandas assert_frame_equals with a rtol and atol tolerance.

**Full Traceback**
https://github.com/NREL/reV/pull/289/checks?check_run_id=2271778550

**Code Sample**
See PV and wind tests:
https://github.com/NREL/reV/blob/main/tests/test_gen_wind.py
https://github.com/NREL/reV/blob/main/tests/test_gen_pv.py

**Expected behavior**
More deterministic test results with PySAM version changes

**Charge code**
WEPG 10348 08.01.01

mikebannis:
That one has bit me already. I'm unsure on the difference between hard-coded and caching in the context above. Should the tests be generating PySAM 2.1.4 results on the fly or should they be saved in the repo?

MRossol:
Save them in the repo:
https://github.com/NREL/reV/blob/719b3ed16c72bfed4eccde50507cf0dd89dae1db/tests/test_gen_pv.py#L88-L89

MRossol:
@mikebannis Anything I can do here to close this out and merge the #289

mikebannis:
@MRossol I haven't gotten to this yet. Do you have a timeline?

MRossol:
I'll take a wack at it while you focus on Transmission costs

Issue #273, opened 2021-02-24
Integrate PySAM 2.2.1


grantbuster:
**Why this feature is necessary:**
Need to integrate the [new PySAM version](https://github.com/NREL/ssc/pull/537) that fixes bugs related to leap year timeseries. Can now update the [reV-PySAM requirement](https://github.com/NREL/reV/blob/a6ea5869de98856207722f06de59f2876a0edcfb/requirements.txt#L1) to be >=2.2.1. I think we'll need to relax some the assert precision on some of the solar thermal tests that are seeing minor differences (fractions of a percent).

grantbuster:
Future feature might be to run reV-gen with sub-annual (daily or weekly) input resource.

PR #298, opened 2021-04-21
Pysam update

MRossol:
- Fix CSP, SWH, Linear, Trough tests to use a baseline file and np.allclose
- Remove dirout from Gen/Econ and use out_fpath instead


MRossol:
closes #290

MRossol:
closes #273

grantbuster (approved):

PR #303, opened 2021-05-11
Gb/gid map

grantbuster:


reV/SAM/SAM.py
```diff
@@ -65,7 +65,37 @@ def _get_base_handler(res_file, module):
...
+    def _parse_gid_map_sites(gen_gids, gid_map):
+        """Parse resource gids based on the generation gids used by
+        project_points and a gid_map. If gid_map is None, the input gen_gids
+        are just passed through as the res_gids.
+
+        Parameters
+        ----------
+        gen_gids : list
+            List of project_points "sites" that are the generation gids.
+        gid_map : None | dict
```
MRossol, line 77: Do we want to set this as a kwarg with default None? I find that to be more intuative than having the possible dtypes be None | dict

reV/generation/generation.py
```diff
@@ -265,6 +287,69 @@ def run(cls, points_control, tech=None, res_file=None, output_request=None,
 
         return out
 
+    def _parse_gid_map(self, gid_map):
```
MRossol, line 290: Should this go in GenBase?

reV/SAM/SAM.py
grantbuster: Yeah that might be easier to read. The only reason i didnt is because its a private method that should always be called with both arguments. Calling the method without gid_map is pointless so i wanted to make sure it gets input.

reV/generation/generation.py
grantbuster: I don't think so... I didnt really see any reason to have this available in the econ module. There's not a real use case I can come up with and it gets super confusing because there would be res_gids -> gen_gids -> econ_gids.

MRossol: That was my suspicion and works for me

MRossol (dismissed):

reV/SAM/SAM.py
grantbuster: Yep agreed on the readability. will do!

PR #306, opened 2021-05-18
Gb/sam tz input

grantbuster:
Minor feature to better allow for timezone and elevation to be input via the reV gen/econ site_data input. Useful feature when we're dealing with forecast or other misc datasets with weird meta data.

### DIFF
--- a/reV/SAM/generation.py
+++ b/reV/SAM/generation.py
@@ -112,13 +112,16 @@ class Generation(RevPySam, ABC):
         return res_mean, out_req_nomeans
 
     @staticmethod
-    def tz_check(sam_sys_inputs, meta):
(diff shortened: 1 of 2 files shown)


MRossol (approved):
Looks good, can we add a test?

grantbuster:
Yes can add test... any ideas on what a good test would be?

MRossol:
I think a baseline forecast test that uses the gid_map and site_data with tz and elevation data would be great, both from an example standpoint and just to make sure both functions are passing data through properly

grantbuster:
Okay added kind of an integrated gen forecast test that we can pile our misc forecast features into. I think this works!

Issue #310, opened 2021-05-25
Pass through the capital cost multiplier to SC_Table


MRossol:
**Why this feature is necessary:**
Needed for ReEDS

**A possible solution is:**
Compute it during economies of scale and add to meta data

**I have considered the following alternatives:**

**Additional context**

**Charge code**
WEPG 10348 08.01.01

**Urgency / Timeframe**
ASAP

PR #312, opened 2021-05-25
passed through economies of scale capital cost scaling value to sc ag…

grantbuster:
…g table (with tests)


grantbuster:
implements issue #310

MRossol (approved):
BOOM

Issue #311, opened 2021-05-25
Supply Curve v2


MRossol:
**Why this feature is necessary:**
Update Supply Curve to take in the new least-cost transmission tables with pre-computed costs. New table will only do the sorting.

For Offshore: Use least cost tables, map offshore sc point to nearest onshore sc point. Run simple sorting to connect to nearest transmission feature

**A possible solution is:**

**I have considered the following alternatives:**

**Additional context**

**Charge code**
If you are at NREL, please provide a task number for the developers to implement this feature.

**Urgency / Timeframe**
When do you need this feature by? How urgent is it related to other feature requests you've made?

MRossol:
Implemented in #314

PR #314, opened 2021-06-14
Sc update

MRossol:
Was actually pretty straightforward. Only had to make two changes:
1) If trans_cap_cost in trans_table don't compute costs
2) Recursion magic to merge sc_points and trans_tables by capacity!

Also some general clean-up.

Still need to test/write tests, waiting on data from @mikebannis


MRossol:
closes #311

reV/supply_curve/supply_curve.py
```diff
@@ -256,30 +269,51 @@ def _merge_sc_trans_tables(cls, sc_points, trans_table,
...
+                trans_sc_table = []
+                for cap, table in trans_table.groupby('max_cap'):
+                    mask = sc_points['capacity'] <= cap
+                    trans_sc_table.append(cls._merge_sc_trans_tables(
+                        sc_points.loc[mask].copy(),
+                        table,
+                        compare_cap=False,
+                        sc_cols=sc_cols))
+
+                trans_sc_table = pd.concat(trans_sc_table)
```
grantbuster, line 292: regarding syntax and read-ability, I tend to think this spacing method isn't appropriate here. I understand that you want an empty space after for-loops, but having the pd.concat() call directly above the else: statement visually joins the two, while in fact it would be more intuitive to see the pd.concat() grouped with the for-loop instead of the else: indent level. Does that make sense? If you like it this way don't feel obligated to change it, but I just dont believe in "absolute" rules in cases like this.

reV/supply_curve/supply_curve.py
```diff
@@ -256,30 +269,51 @@ def _merge_sc_trans_tables(cls, sc_points, trans_table,
...
+
+            if compare_cap and 'max_cap' in trans_table:
+                trans_sc_table = []
+                for cap, table in trans_table.groupby('max_cap'):
+                    mask = sc_points['capacity'] <= cap
+                    trans_sc_table.append(cls._merge_sc_trans_tables(
+                        sc_points.loc[mask].copy(),
+                        table,
+                        compare_cap=False,
+                        sc_cols=sc_cols))
```
grantbuster, line 290: I think i understand what this is doing. Is it okay if there are duplicates afterward? Is there any check for duplicates if not?

grantbuster (changes requested):

reV/handlers/transmission.py
```diff
@@ -51,7 +51,7 @@ def __init__(self, trans_table, line_tie_in_cost=14000, line_cost=3667,
 
         logger.debug('Trans table input: {}'.format(trans_table))
         logger.debug('Line tie in cost: {} $/MW'.format(line_tie_in_cost))
-        logger.debug('Line cost: {} $/MW-mile'.format(line_cost))
+        logger.debug('Line cost: {} $/MW-km'.format(line_cost))
```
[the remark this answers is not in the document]
MRossol: Yup! At least thats the hope plan...

reV/handlers/transmission.py
```diff
@@ -148,6 +148,14 @@ def _parse_table(trans_table):
             logger.error(ex)
             raise
 
+        trans_table = \
+            trans_table.rename(columns={'trans_line_gid': 'trans_gid',
+                                        'trans_gids': 'trans_line_gids'})
```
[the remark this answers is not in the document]
MRossol: So originally the naming was as follows which IMO made no sense:
`trans_line_gid` == the transmission feature id, No idea why "line" was in the name
I've had Bannister re-name this column to `trans_gid` in the new tables

`trans_gids` == the list of transmission line gids that connect to the feature (Substations only). No idea why line wasn't in this name so I added it in. Also wanted to make it less ambiguous with respect to `trans_gid`

reV/supply_curve/supply_curve.py
```diff
@@ -49,7 +49,7 @@ class SupplyCurve:
         Capital cost of connecting each supply curve point to their respective
         transmission feature. This value includes line cost with
         transmission_multiplier and the tie-in cost. Default units are $/MW.
-    dist_mi : float
+    dist_km : float
         Distance in miles from supply curve point to transmission connection.
```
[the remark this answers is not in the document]
MRossol: dop! Will do

reV/supply_curve/supply_curve.py
```diff
@@ -68,9 +68,10 @@ def __init__(self, sc_points, trans_table, fcr, sc_features=None,
         sc_points : str | pandas.DataFrame
             Path to .csv or .json or DataFrame containing supply curve
             point summary
-        trans_table : str | pandas.DataFrame
+        trans_table : str | pandas.DataFrame | list
             Path to .csv or .json or DataFrame containing supply curve
-            transmission mapping
+            transmission mapping, can also be a list of transmission tables
+            with different line voltage (capacity) ratings.
```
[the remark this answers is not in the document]
MRossol: Good call I'll add that in

reV/supply_curve/supply_curve.py
```diff
@@ -244,7 +253,11 @@ def _merge_sc_trans_tables(cls, sc_points, trans_table,
             Table of supply curve point summary
         trans_table : pd.DataFrame | str
             Table mapping supply curve points to transmission features
-            (either str filepath to table file or pre-loaded dataframe).
+            (either str filepath to table file, list of filepaths to tables by
+             line voltage (capacity) or pre-loaded dataframe).
+        compare_cap : bool, optional
+            Flag to compare the sc_point and transmission feature capacities,
+            'capacity' <= 'max_cap'
```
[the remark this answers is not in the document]
MRossol: So this came up during the recursion. Originally I was going to drop the min/max_cap columns from the trans_tables but figured it would be good to keep those for debugging. The flag isn't exposed to the user its only used inside the method during the recursion. Its on by default but turned off here:  https://github.com/NREL/reV/blob/c42b431184d6f8ce9ea3bcb50472d9cbe45d3955/reV/supply_curve/supply_curve.py#L289

when we are iterating through the different capacity "bins" This is to make sure we don't end up in a recursion loop

reV/supply_curve/supply_curve.py
```diff
@@ -256,30 +269,51 @@ def _merge_sc_trans_tables(cls, sc_points, trans_table,
             This is performed by merging left with trans_table, so there may be
             rows with nan sc_gid.
         """
-        trans_table = cls._parse_trans_table(trans_table)
+        if isinstance(trans_table, (list, tuple)):
+            trans_sc_table = []
+            for table in trans_table:
+                trans_sc_table.append(cls._merge_sc_trans_tables(
```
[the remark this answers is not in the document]
MRossol: See earlier comment

reV/supply_curve/supply_curve.py
MRossol: I think thats fair and a good "exception"

reV/supply_curve/supply_curve.py
MRossol: Good catch, this is what I realized when we were chatting earlier. I've had Bannister add in a min_cap column and updated the code to make the bins non-overlapping

reV/handlers/transmission.py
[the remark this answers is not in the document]
grantbuster: Gotcha and totally agree. Can you add an in-line comment describing the feature gid vs. line-gids (for substations)? This is going to appear very confusing as soon as next week.

reV/supply_curve/supply_curve.py
mikebannis: Is there an easy way to reduce the total amount of indentation? I'm looking at this with a lot of github comments in it, but it's hard to see what lines up with what.

grantbuster: I would almost always suggest having the code side by side with the github markup. The github sequential markup (instead of side by side) is really difficult to read. 

As far as reducing indentation: write more functions!

MRossol: And ruin my recursion!?!?!

reV/handlers/transmission.py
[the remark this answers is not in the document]
MRossol: # Update legacy transmission table columns to match new less ambiguous
        # column names:
        # trans_gid -> the transmission feature id, legacy name: trans_line_gid
        # trans_line_gids -> gids of transmission lines connected to the given
        # transmission feature (only used for Substations),
        # legacy name: trans_gids

reV/supply_curve/supply_curve.py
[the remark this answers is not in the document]
grantbuster: hmmm okay. the reason for this still kind of confuses me. the goal here is to join transmission tables with SC agg tables for different max capacities? Could we also do a left join and drop duplicates based on the capacity comparison? That would reduce one confusing level of recursion and also make a natural location to check for duplicates.

reV/supply_curve/supply_curve.py
grantbuster: haha no no, just general comment about lots of indentation

reV/supply_curve/supply_curve.py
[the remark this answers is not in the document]
MRossol: I don't follow how that would work. The inputs are as follows:
sc_gids of interest
each trans_table contains all sc_points and their connection to the transmission features in the given bin (> min_cap, <= max_cap)

We want to pull the transmission feature from the proper table for each sc_gid based on the sc_gids capacity. I don't follow how we can properly delete duplicates. Do we just join the tables and then delete duplicate trans_gids based on the transmission capacity (retaining the smallest capacity feature?)

MRossol:
@grantbuster I think this should be ready for Travis. I added tests to check against a baseline and to check to make sure that the final sc table is connecting sc_points to features in the correct capacity bins.

One NOTE for Travis, I updated the trans_cap_cost column to be trans_cap_cost_mw. This is to help distinguish the costs coming in from Mike and the costs computed in reV based off of the sc_gid capacity.

grantbuster:
Okay, looks like tests are still failing so lets let travis run off of a dev branch. 

is `trans_cap_cost_mw` a dollars per MW unit? I would suggest including the `per_mw`, as a numerator/denominator confusion could be disastrous.

MRossol:
I'll fix the tests now, but good call. I'll change it to _per_mw

reV/supply_curve/supply_curve.py
```diff
@@ -855,12 +855,6 @@ def full_sort(self, sort_on='total_lcoe',
                                        sort_on=sort_on, columns=columns,
                                        downwind=downwind)
 
-        if 'offshore' in supply_curve:
-            sum_cols = {'combined_cap_cost': ['array_cable_CAPEX',
-                                              'export_cable_CAPEX',
-                                              'trans_cap_cost_per_mw']}
```
grantbuster: @MRossol fyi i think we talked about this previously but i remember what this was for. It was to sum up some of the cabling costs to get an idea of the offshore transmission capex. You can do most of this in NRWAL now and manually sum some columns in the SC table otherwise. I think it's safe to remove and the hard-coded column names were causing travis problems.

MRossol: Sounds good to me and glad Travis' bug wasn't "me"

grantbuster: I wasnt blaming you but i think its good to remove the hard-coded col names nonetheless

MRossol: I wasn't saying you were, I was just worried i'd fucked up seeing as i pushed that at 10pm last night and the first thing I saw when logging on was, "it failed"...

grantbuster: oh haha yeah gotcha, might have misread. yeah the tests are still failing but thats unrelated i think.

grantbuster (commented):
What was the cause of the discrepancy on this?

MRossol:
One the baseline tests? I'm 99% sure the difference was in the file ordering when multiple tables are loaded. I noticed that there are a few transmission features with identical cost and the new tests were connecting to a different feature. Seems like it was a strange sort change.

PR #317, opened 2021-07-09
Gb/detailed pv

grantbuster:


reV/SAM/econ.py
```diff
@@ -509,19 +506,15 @@ def _windbos(inputs):
                         outputs = wb.output
         return inputs, outputs
 
-    @property
-    def default(self):
+    @staticmethod
```
MRossol, line 509: I'm confused by this change, Is default now just a place holder?

reV/SAM/generation.py
```diff
@@ -34,6 +39,56 @@
 class Generation(RevPySam, ABC):
```
MRossol, line 39: Do we want to rename this BaseGeneration since its an abstract class?

reV/SAM/generation.py
```diff
@@ -371,128 +436,49 @@ def reV_run(cls, points_control, res_file, site_df,
 
 
 class Solar(Generation, ABC):
```
MRossol, line 438: Same here.. I keep thinking that these are user classes not base/abstact classes

reV/SAM/generation.py
```diff
@@ -537,22 +523,30 @@ def set_nsrdb(self, resource):
         self['solar_resource_data'] = resource
 
 
-class Pvwatts(Solar, ABC):
-    """Photovoltaic (PV) generation with pvwattsv5.
+class AbstractPv(Solar, ABC):
```
MRossol, line 526: why AbstactPv instead of AbstractPV, I know its not perfect. camel case but Pv  is weird...

reV/SAM/generation.py
```diff
@@ -1067,18 +1022,15 @@ def cf_mean(self):
 
         return net_power / name_plate
 
-    @property
-    def default(self):
+    @staticmethod
```
MRossol, line 1025: I still don't get this change. isn't it less efficient to recompute the defaults over and over? Why not add another CLS variable called `DEFAULT_CLS` and then have this:
```
@property
def default(self):
    if self._default is None:
        self._default = self.DEFAULT_CLS.default()
        
        
     return self._default

**Review by MRossol — CHANGES_REQUESTED** [review]
I'm confused by the change for defaults from a property to a staticmethod. I remember seeing a commit about this, but mind explaining?

Also I found the SAM/generation.py class names a bit hard to follow:
1) `Generation` and `Solar` are abstract/base classes like `AbstractPv` and I think should be renamed as such
2) I know Pv is proper camelcase but I found it weird. I feel like `AbstractPV` is less ambiguous. Same with `Pvwatts*`, Why not `PVwatts*` or `PvWatts*`. This one is a bit silly, but food for thought

File: reV/SAM/SAM.py [hunk 6]
```diff
@@ -565,6 +562,8 @@ def __init__(self, meta, sam_sys_inputs, output_request,
         self.sam_sys_inputs = sam_sys_inputs
         self.site_sys_inputs = site_sys_inputs
         self.output_request = output_request
+        if self.output_request is None:
```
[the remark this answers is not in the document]
grantbuster: lol i really dont have a strong opinion. The way it is with `self.` has cleaner indentation. Too similar to the point of apathy though imo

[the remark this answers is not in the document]
MRossol: Agreed on the apathy, definitely a 6 and 1/2 a dozen situation, I was just curious. Indentation is as good as answer as any!

reV/SAM/econ.py
grantbuster: Placeholder? no, each concrete reV-SAM object will have a default() method to retrieve the default object. Our `Sam` wrapper has an abstractmethod placeholder but all the concrete classes will have useful methods.

Previously it was a property which was kind of nice and natural to call, but the problem was that you had to instantiate the object before getting the default object. I think a static method is much better, because now we can explicitly require the resource, meta, and config inputs separately from a default object.

reV/SAM/generation.py
grantbuster: agree, will rename AbstractSamGeneration

reV/SAM/generation.py
grantbuster: agreed

reV/SAM/generation.py
grantbuster: Pv is weird but almost all of these classes have abbreviations in their name anyway... I would rather be rigorous about camel case vs. choosing what abbreviations deserve to be camel case. 

On the same note, i really regret having reV and SAM "properly" capitalized everywhere. I think we should either adhere to python cap convention or not at all. but those changes are probably too big to be really worth it right now.

reV/SAM/generation.py
grantbuster: The only place we use the defaults is in the code below, and we retrieve the defaults once and save them, so we dont end up retrieving them again and again anyway (im pretty sure). 

The solution you proposed would still require the class the be instantiated to retrieve the default, which is what i was trying to move away from.

https://github.com/NREL/reV/blob/95082ea79663eff04259cfd9d98b022e60b158d8/reV/generation/base.py#L835-L838

MRossol: Gotcha, this and your earlier comment cleared up my confusion!

reV/SAM/generation.py
MRossol: Thats fair, but on that point can we change the following:
Pvwatts* -> PvWatts*
Pvsam1 -> PvSam1

grantbuster: agreed!

MRossol (approved):

PR #318, opened 2021-07-09
Gb/agg lcoe

grantbuster:


reV/supply_curve/point_summary.py
```diff
@@ -206,8 +215,29 @@ def mean_lcoe(self):
         mean_lcoe : float | None
             Mean LCOE value for the non-excluded data.
         """
+
         mean_lcoe = None
-        if self.lcoe_data is not None:
+
+        # prioritize the calculation of lcoe explicitly
+        # from the multi year mean CF
```
MRossol, line 222: This doesn't necessarily have to be the MY mean CF, maybe just say from the mean CF?

reV/supply_curve/point_summary.py
```diff
@@ -206,8 +215,29 @@ def mean_lcoe(self):
...
+        if self._recalc_lcoe:
+            required = ('fixed_charge_rate', 'capital_cost',
+                        'fixed_operating_cost', 'variable_operating_cost',
+                        'system_capacity')
+            if self.mean_h5_dsets_data is not None:
+                if all(k in self.mean_h5_dsets_data for k in required):
```
MRossol, line 228: Do we want to raise an error if the required datasets are not found?

reV/supply_curve/sc_aggregation.py
```diff
@@ -603,60 +616,68 @@ def _get_input_data(gen, gen_fpath, econ_fpath, res_class_dset,
...
-        if lcoe_dset in gen.datasets:
-            lcoe_data = gen[lcoe_dset]
-        else:
-            lcoe_data = None
-            w = ('Could not find lcoe dataset "{}" in generation file: {} or '
-                 'econ file: {}. Available datasets: {}'
-                 .format(lcoe_dset, gen_fpath, econ_fpath, gen.datasets))
-            logger.warning(w)
-            warn(w, OutputWarning)
+        lcoe_recalc_req = ('fixed_charge_rate', 'capital_cost',
```
MRossol, line 637: I'd think it would help to move all of this logic into its own method. Its a lot and a dedicated docstring explaining the difference between missing, missing1, and missing2 would go a long way.

MRossol (commented):
Damn you weren't kidding about not wanting to maintain this in more than one place!

reV/supply_curve/point_summary.py
grantbuster: we can but there would be no point to re-calculating the LCOE from a single year mean, so i think this comment is appropriate in describing this code block

MRossol: Okay thats what I thought, but I'm still a bit confused, how does the code know that its MY Means and to re-compute? If its doesn't maybe we should just clarify that this code will prioritize the re-calculation of lcoe explicitly in case the input data is a Multi-year mean?

grantbuster: The code doesnt know if its multi year CF or not, but if its single year CF then the re-calculated LCOE should match / not matter. the code here is really of consequence for multi year. Another reason we shouldn't raise errors or force the user into anything.

reV/supply_curve/point_summary.py
grantbuster: There are warnings in the `SupplyCurveAggregation._get_input_data()` method (remember how we talked about warnings/errors in the non-parallel pre-flight checks)

reV/supply_curve/point_summary.py
MRossol: Seeing as this still confuses the hell out of me I still think
1) making the comment clearer, even jus to stat the above would be helpful
2)Somewhere to raise a warning that LCOE can't re recomputed and that if you are running MY mean CF your results might be off, I think you do that below (third comment), but just wanted to make sure.

reV/supply_curve/sc_aggregation.py
grantbuster: I disagree. This is a single purpose method that will only be called once. Everything in the method is in the spirit of a single purpose - to check and retrieve input data. Besides, a method docstring won't describe the internal "missing" variables. I think the real solution here is just to better name the "missing" variables?

MRossol: That works too. I still think you could have a checking method inside the retrieval method, BUT as you said the important part is making the logic easier to follow. Renaming the internal missing variables should accomplish that.

grantbuster: As our code base continues to grow, i think we really need to think harder about the method chaining. This is a great example. It almost hits the complexity limit, but as far as code design goes, it is a totally appropriate single-purpose method. There are some huge benefits to smaller methods with good doc strings, but there's also a huge challenge in chasing an object as it gets passed down through a billion levels of nested methods. As the zen of python says: "Flat is better than nested"

MRossol: I'm super torn on this one. I definitely agree with avoiding unnecessary method chaining, BUT I also get the need to limit complexity. I stumbled upon this article this weekend about [refactoring legacy code](https://opensource.com/article/21/7/legacy-code ) and his complex if statement example is what made me suggest a sub-method here.

MRossol (dismissed):
Looks good, but we somehow skipped version 0.5.5. No need to change it here, but we should change it on main before we release.

reV/supply_curve/sc_aggregation.py
grantbuster: Yes I think neither extreme is appropriate. But here I think we have a good example of narrow purpose method that needs to have a somewhat higher complexity. We could break it up into a bunch of different methods but that would only add depth and i dont think it would add convenience. From that article, we are VERY far from the statement "If you cram all capabilities an object can offer into a single, giant function"

PR #325, opened 2021-09-02
update trans feature to sc point capacity mapping

MRossol:
Fix bug that would drop sc_gids with capacity > max_cap in trans tables


MRossol:
@grantbuster, do you want to review this?

grantbuster:
No thats okay, i dont really understand what the issue is. Why are the reVX tests failing?

MRossol:
no idea, there are no logs but it also says it passed, I'm guessing a github actions hiccup.

Do you mind putting quick eyes on the change. I want to make sure I didn't mess up the logic. The issues was that with the new transmission tables we need to map the SC points to the proper transmission table based on capacity as follows:

`sc_point['capacity'] <= trans_table['max_cap']`

The issue with this that Travis found was that there are some sc_points with capacity > 1500MW which is the capacity of the largest tie-lines. So these features were connected to garbage.

The fix I've hopefully implemented is to take all of the sc_points that exceed the max_cap in the tranmission tables and for each transmission feature that they map to, connect to the feature with the largest max_cap.

reV/supply_curve/supply_curve.py
```diff
@@ -259,6 +259,42 @@ def _parse_trans_table(trans_table):
...
+    def _map_trans_capacity(trans_sc_table):
+        """
+        Map SC gids to transmission features based on capacity. For any SC
+        gids with capacity > the maximum transmission feature capacity, map
+        SC gids to the feature with the largest capacity
+
+        Parameters
+        ----------
+        trans_sc_table : pandas.DataFrame
+            Table mapping supply curve points to transmission features.
```
grantbuster, line 272: So this is after joining the SC agg table to the Trans "all cons" table right?

reV/supply_curve/supply_curve.py
```diff
@@ -259,6 +259,42 @@ def _parse_trans_table(trans_table):
...
+
+        Returns
+        -------
+        trans_sc_table : pandas.DataFrame
+            Updated table mapping supply curve points to transmission features
+            based on maximum capacity
```
grantbuster, line 278: So this will have connection rows removed where SC capacity > trans capacity, unless SC capacity > all trans max capacity.

reV/supply_curve/supply_curve.py
```diff
@@ -259,6 +259,42 @@ def _parse_trans_table(trans_table):
...
+            msg = ("The following SC points have a capacity that "
+                   "exceeds the maximum transmission feature capacity "
+                   "{}, and will be mapped to features with the max "
+                   "capacity:\n{}"
+                   .format(over_max['sc_gid'].unique(),
+                           np.max(over_max['max_cap'])))
+            logger.warning(msg)
+            warn(msg)
+            for _, grp_df in over_max.groupby(['sc_gid', 'trans_gid']):
+                idx_max = grp_df['max_cap'].argmax()
```
grantbuster, line 293: max_cap is the transmission feature maximum capacity?

reV/supply_curve/supply_curve.py
```diff
@@ -259,6 +259,42 @@ def _parse_trans_table(trans_table):
...
+                trans_sc_table = trans_sc_table.append(grp_df.iloc[[idx_max]])
```
grantbuster, line 294: I think there's an unnecessary amount of bracketing going on here.

reV/supply_curve/supply_curve.py
```diff
@@ -279,7 +315,7 @@ def _merge_sc_trans_tables(cls, sc_points, trans_table,
 
         Returns
         -------
-        trans_table : pd.DataFrame
+        trans_sc_table : pd.DataFrame
             Updated table mapping supply curve points to transmission features.
             This is performed by merging left with trans_table, so there may be
             rows with nan sc_gid.
```
grantbuster, line 321: what?? why would there be nan sc gid? When you say "merge left with trans_table" (ambiguous!), you mean the trans_table is the left table and how=left?

grantbuster (changes requested):
Logic looks great but minor comments and questions

reV/supply_curve/supply_curve.py
grantbuster: I also think we should move the call `self._check_sc_trans_table(self._sc_points, trans_table)` into the `_merge_sc_trans_tables()` method

reV/supply_curve/supply_curve.py
MRossol: yes, this is the fully agg_table + trans_features table merged on sc_point_gid

reV/supply_curve/supply_curve.py
MRossol: yes, thats the hope...

reV/supply_curve/supply_curve.py
MRossol: its the maximum capacity of the tie-lines connecting the sc_point to that features

reV/supply_curve/supply_curve.py
MRossol: You'd think right! turns out you can't append a series to a dataframe. Doing iloc[[]] returns a dataframe

reV/supply_curve/supply_curve.py
MRossol: hmm, good point,  that specific code is years old. I'll double check

sounds good on the check, i'll move that

reV/supply_curve/supply_curve.py
grantbuster: I didnt know that! cool.

grantbuster (approved):

PR #326, opened 2021-09-13
added a few additional supply curve logging statements

grantbuster:


reV/handlers/transmission.py
```diff
@@ -279,23 +279,31 @@ def _calc_cost(distance, line_cost=2279, tie_in_cost=0,
...
 
-
         Returns
         -------
         avail_cap : float
             Substation available capacity
         """
 
-        line_caps = [self[gid]['avail_cap'] for gid in line_gids]
+        try:
```
MRossol, line 299: Nice, surprised this didn't bite us earlier...

MRossol (approved):

reV/handlers/transmission.py
grantbuster: it's one of those "fine when it works" type things haha

PR #327, opened 2021-09-13
moved pre extract exclusion mask method to exclusion handler

grantbuster:


MRossol (approved):
looks good, nice clean-up

PR #331, opened 2021-09-21
add recalc_lcoe flag to sc agg cli and config

grantbuster:


MRossol (approved):
The CLIs have become so much cleaner!

Not necessary but we do want to add a CLI test to aggregation?

grantbuster:
I mean yes i'd never say no to more tests haha

PR #333, opened 2021-10-04
Gb/wave nrwal

grantbuster:


reV/offshore/offshore.py
```diff
@@ -441,49 +461,71 @@ def run_nrwal(self):
...
+                f._add_dset('lcoe_fcr', lcoe, np.float32,
+                            attrs={'units': 'dol/MWh', 'scale_factor': 1})
+
+            f._add_dset('cf_mean_raw', self._cf_mean, f.dtypes['cf_mean'],
+                        attrs=f.attrs['cf_mean'])
+            self._cf_mean[self._offshore_mask] *= loss_mult
+            f['cf_mean'] = self._cf_mean
+
+            profiles = None
+            if 'cf_profile_raw' in f.dsets:
```
MRossol, line 522: what is the difference between cf_profile_raw and cf_profile

MRossol (dismissed):
One question, but looks good. Nice refactors along the way!

reV/offshore/offshore.py
grantbuster: the _raw datsets are CF before NRWAL losses are applied. We save the raw datasets so we can rerun NRWAL and not doubly count losses.

MRossol: Sweet, that was my guess!

PR #337, opened 2021-11-09
Gb/basic agg

grantbuster:


grantbuster:
Reorganization of aggregation classes and addition of basic aggregation without generation file to the sc-agg CLI

PR #343, opened 2022-01-18
pp/rep_profiles_refactor

ppinchuk:
Fixed bug causing program to crash when `ggids=None`. Removed redundant `AggregatedRepProfiles` class in favor of `RepProfiles`. Updated cli to use `RepProfiles`. Made `reg_cols` a required cli argument so that cli would error out if that option is not provided during invocation and before submitting to eagle. Marked `reg_cols` as a requirement in `RepProfilesConfig`.


grantbuster (approved):
This looks great! Removing (my) garbage code is just as important as contributing new code.

PR #348, opened 2022-02-15
NRWAL constant equation output request fix

ppinchuk:
The issue:
Currently, an implicit assumption is made that when a user asks for a "calculation" output from NRWAL, it will be in the form of an array corresponding to each site. This is likely true for most cases, but it breaks in some edge cases. For example, suppose the user has a lease price input in their config file. Suppose they also want this value scaled to  millions of dollars, so they add a `lease_price_mil: lease_price / 1000000` calculation. If the user wants to request this value as an output (e.g. for debugging purposes), rev-NRWAL assumes it comes in the form of an array, where in reality the output is just a float.

The solution:
Before saving any NRWAL calculation output, single values are cast to numpy arrays. Implemented in this PR.

### DIFF
--- a/reV/nrwal/nrwal.py
+++ b/reV/nrwal/nrwal.py
@@ -569,7 +569,7 @@ class RevNrwal:
     def _value_to_array(self, value, name):
         """Turn the input into numpy array if it isn't already."""
         if np.issubdtype(type(value), np.number):
-            value *= np.ones(len(self._site_data))
+            value *= np.ones(len(self.analysis_gids), dtype=np.float32)
 
         if not isinstance(value, np.ndarray):
             msg = ('NRWAL key "{}" returned bad type of "{}", needs to be '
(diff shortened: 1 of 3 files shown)


reV/nrwal/nrwal.py
```diff
@@ -561,6 +563,11 @@ def _save_nrwal_misc(self, name, nrwal_config, output_mask):
             assert not any(value.variables), msg
             value = value.eval()
 
+        value = self._value_to_array(value, name)
+        self._out[name][output_mask] = value[output_mask]
+
+    def _value_to_array(self, value, name):
+        """Turn the input into numpy array if it isn't already."""
         if np.issubdtype(type(value), np.number):
             value *= np.ones(len(self._site_data))
```
grantbuster, line 572: I realize that i wrote this but can you change it to:

value *= np.ones(len(self.analysis_gids), dtype=np.float32)

grantbuster (changes requested):
Can you also remove any (commented) breakpoint() calls? I think i left one in there on accident.

grantbuster (approved):

reV/nrwal/nrwal.py
ppinchuk: Done and done!

PR #350, opened 2022-02-28
Added hybrids module

ppinchuk:
Added new hybrids module to facilitate the hybridization of representative profiles from wind and solar generation


reV/hybrids/hybrids.py
```diff
@@ -0,0 +1,1170 @@
...
+            `ratios={RatioColumns('solar_capacity', 'wind_capacity'):
+            (0.5, 1.5)}` would adjust both the solar and wind capacities
+            such that the solar and wind capacity ratio is always between half
+            and double (e.g., no capacity value would be more than double the
+            other). Finally,
+            `ratios={RatioColumns('solar_capacity', 'wind_capacity',
+            fixed='wind_capacity'): (0.5, 1.5)}` would limit the solar and
+            wind capacities to be between half and double **without altering
+            the value of the wind_capacity** (e.g. only the solar capacity
+            would be adjusted). By default, None (no limit).
```
grantbuster, line 380: A few comments:

1. Something this complicated may deserve its own class. In argument docstrings for `ratios` you could just say "Option to set ratios... see the HybridsRatio class for details". You could have the class be a general HybridLimits class that does the logic for both limits and ratio limits. 
2. Having different specification between the CLI and the python class is going to be very difficult to maintain (looks like you've already spent quite a lot of time writing the separate doc strings). I would recommend defining a single format that is CLI-friendly.
3. This is a dictionary but i expect (hope) that you cannot set ratio bounds on more than a single pair of columns? I feel like the logic would get pretty messy with ratio bounds on capacity AND something else (no idea why you would want to do this anyway). 
4. Possible options to resolve bullets 2 and 3 above could be argument format like: `[[ratio_col_numerator, ratio_col_denominator], [ratio1, (optional)ratio2]]` or two arguments like: `ratio_cols=[ratio_col_numerator, ratio_col_denominator], ratio_bounds=(0, 1)` I actually think the 2x arguments would be clean and would be more cli-friendly in the long run even if you do have to add another argument.

reV/hybrids/hybrids.py
```diff
@@ -0,0 +1,1170 @@
...
+        """
+
+        for cols in self._ratios:
+            if not isinstance(cols, RatioColumns):
+                msg = ("Key {!r} for input 'ratios' not understood. "
+                       "Please make sure this value is an instance of "
+                       "`RatioColumns`")
+                e = msg.format(cols)
+                logger.error(e)
+                raise InputError(e) from None
```
grantbuster, line 490: I dont think you need the `from None` here? I thought you just used `raise X from E` when you're catching a different exception E

tests/test_hybrids.py
```diff
@@ -0,0 +1,865 @@
+# -*- coding: utf-8 -*-
+"""reV hybrids tests.
```
grantbuster, line 2: This is incredibly thorough. Nice work!!!

tests/test_hybrids.py
```diff
@@ -0,0 +1,865 @@
...
+
+    limits = {'solar_capacity': 100, 'wind_capacity': 0.5}
+
+    h = Hybridization(SOLAR_FPATH, WIND_FPATH, limits=limits).run()
+
+    assert np.all(h.hybrid_meta['solar_capacity'] <= limits['solar_capacity'])
+    assert np.all(h.hybrid_meta['wind_capacity'] <= limits['wind_capacity'])
+
+
+def test_invalid_limits_column_name():
```
grantbuster, line 284: Love all the negative tests

tests/test_hybrids.py
```diff
@@ -0,0 +1,865 @@
...
+    ((0.3, 3.6), (0.3 - 1e6, 3.6 + 1e6))
+])
+def test_ratios_input(ratio_cols, ratio, bounds):
+    """Test that the hybrid meta limits the ratio columns correctly. """
+    ratio_cols = RatioColumns(*ratio_cols)
+    if ratio_cols.fixed is None:
+        h = Hybridization(
+            SOLAR_FPATH, WIND_FPATH,
+            ratios={ratio_cols: ratio}
+        ).run()
```
grantbuster, line 222: Since you're already running the profile hybridization, maybe check to make sure that the profiles never exceed the ratio limits too? Or just check to make sure the profiles never exceed the stated hybrid capacity in the meta data.

grantbuster (changes requested):
Overall a simply excellent job.

reV/hybrids/hybrid_methods.py
```diff
@@ -0,0 +1,176 @@
...
+    >>> assert 'scaled_elevation' in h.hybrid_meta.columns
+
+    """
+    def _register(func):
+        HYBRID_METHODS[col_name] = func
+        return func
+    return _register
+
+
+@hybrid_col('hybrid_solar_capacity')
```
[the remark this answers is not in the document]
ppinchuk: Yea, when writing this, I thought more people would be using these objects in python. In that case, the decorator makes it easy to extend the functionality. Now it seems like that's not really necessary, so I will remove this implementation in favor of your suggestion.

reV/hybrids/hybrids.py
grantbuster: After conversation: fixed is unnecessary, you can set bounds that will build full capacity of either wind or solar. One good option for args is `ratio_cols="wind_col/solar_col", ratio_bounds=(bound1, bound2)` and then just assert a single `/` char is in the ratio_cols string and do the min/max on ratio_bounds to get the ratio min/max bounds

reV/hybrids/hybrids.py
```diff
@@ -0,0 +1,1170 @@
...
+        """Get the arrays of the hybridized representative profiles.
+
+        Returns
+        -------
+        profiles : dict
+            Dict of hybridized representative profiles.
+        """
+        return self._profiles
+
+    def run(self, fout=None, save_hybrid_meta=True):
```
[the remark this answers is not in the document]
ppinchuk: Ok, I went with the latter, hope that's ok!

reV/hybrids/hybrids.py
```diff
@@ -0,0 +1,1170 @@
...
+        return out
+
+    def _init_profiles(self):
+        """Initialize the output rep profiles attribute."""
+        self._profiles = {
+            k: np.zeros((len(self.hybrid_time_index), len(self.hybrid_meta)),
+                        dtype=np.float32)
+            for k in OUTPUT_PROFILE_NAMES}
+
+    def _run(self):
```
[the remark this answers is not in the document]
ppinchuk: Done with the previous comment

reV/hybrids/hybrids.py
ppinchuk: Implemented refactor as mentioned in second comment. Used `ratio` instead of `ratio_cols`.

tests/test_hybrids.py
ppinchuk: Good idea! Calculating the ratios of the profiles themselves is a bit tricky because there are 0 values that would have to be dealt with (not impossible but annoying). I decided to take the latter suggestion instead.

grantbuster (approved):

grantbuster:
Looks good! Feel free to merge and release a new reV version. Let's just do a minor version since we'll do an intermediate one soon with bespoke.

PR #351, opened 2022-03-01
Some minor fixes to make bespoke cli run from config

ppinchuk:


ppinchuk:
Well that's strange... reVX tests are failing?

reV/bespoke/bespoke.py
```diff
@@ -962,16 +962,21 @@ def _parse_points(excl_fpath, res_fpath, tm_dset, resolution,
...
+            if not os.path.exists(path):
+                raise FileNotFoundError(
+                    'Could not find required exclusions file: '
+                    '{}'.format(path))
+
+            with h5py.File(path, 'r') as f:
+                if self._tm_dset not in f:
+                    raise FileInputError('Could not find techmap dataset "{}" '
+                                         'in exclusions file: {}'
+                                         .format(self._tm_dset, path))
```
grantbuster, line 979: Okay so i'm pretty sure that tm_dset only needs to be on ONE of the excl file paths. Can you try the following? This is how the exclusions should be handled. 

```
from reV.handlers.exclusions import ExclusionLayers
with ExclusionLayers(paths) as excl:
    if self._tm_dset not in excl:
        msg = '...'
        logger.error(msg)
        raise FileInputError(msg)

File: reV/bespoke/cli_bespoke.py [hunk 2]
```diff
@@ -130,9 +130,11 @@ def from_config(ctx, config_file, points_range, verbose):
 
     is_multi_node = (config.execution_control.option in ('eagle', 'slurm')
                      and config.points_range is None)
-    workers = config.execution_control.nodes if is_multi_node else 1
+    workers = config.execution_control.max_workers if is_multi_node else 1
```
grantbuster, line 133: I actually think this was correct before. Do you have a reason for changing this besides the (shitty) naming convention? I can explain the parallelization scheme if you want.

grantbuster (changes requested):

grantbuster:
don't worry about the revx tests. That's a change that i implemented somewhere on a reV method, maybe in the bespoke branch. I pushed a fix to revx so hopefully it'll pass next time.

reV/bespoke/bespoke.py
ppinchuk: Oh great point, this is just an artefact of me still learning the finer details. I will add that change!

reV/bespoke/cli_bespoke.py
ppinchuk: Yea, I have the following exec control in my config file:
```"execution_control": {
	...
	"max_workers": 36,
	"nodes": 1,
	"memory": 79
	...
}```
So I was asking for 36 threads on 1 cpu, and with the code as it was, it was only opening 1 thread on 1 cpu. Are you suggesting that I should instead request 1 thread on 36 nodes? Perhaps an explanation of the parallelization scheme would help.

Comment (ppinchuk) [reply reV/bespoke/cli_bespoke.py:133]: Ah, taking a closer look at the run logs, I see that even with this change, I checked out 36 nodes. SO it seems I really should be using 
```
"execution_control": {
	...
	"nodes": 36,
	"memory": 79
	...
}
```
Is there any point where the `max_workers` input is actually useful?

grantbuster: yeah so max_workers defaults to None which uses all available CPU's on a node. If you are running out of memory you'll reduce max_workers or if you're debugging you'll set max_workers=1 for serial and more coherent logging. 

I think max_workers is fine in the rest of the CLI. You only need the nodes argument in that one spot and then the CLI has to pass max_workers down all the way to the Bespoke class where the process pool starts on a single node.

grantbuster (approved):

PR #352, opened 2022-03-04
CLI updates

ppinchuk:
Major changelist:
- Removed name option from config files. Deprecation warning is now raised if user provides a `"name"` in config, and the provided value is ignored.
- Removed the ability to specify an output directory. Output directory now set to the config directory. `"directories"` block in config file removed, and all directories (e.g. `"log_directory"`, `"collect_directory"`, etc.) must be specified at the top level of the config file. 
- Added enum to hold module names (which must match the module cli commands) in a single place, so that any updates propagate throughout the rest of the code.
- Fixed `unboundlocalerror` in slurm submissions.


reV/pipeline/pipeline.py
```diff
@@ -354,10 +343,10 @@ def _get_cmd(cls, command, f_config, verbose=False):
...
             Python reV CLI call string.
         """
-        if command not in cls.COMMANDS:
-            raise KeyError('Could not recongize command "{}". '
-                           'Available commands are: {}'
-                           .format(command, cls.COMMANDS))
+        if command not in ModuleName.all_names():
+            raise KeyError("Could not recognize command {!r}. "
+                           "Available commands are: {!r}"
+                           .format(command, ModuleName.all_names())) from None
```
grantbuster, line 349: Can we actually define `cls.COMMANDS=ModuleName.all_names()` to maintain the class attribute functionality? We use this in NSRDB.

grantbuster (changes requested):
Hey this looks good, but there are a lot of complex changes and we dont have good cli tests. Have you tested this by running at least part of a pipeline on Eagle from the new config setup?

One requested change and can you also update the config files in the examples directory?

reV/pipeline/pipeline.py
ppinchuk: Yup! I will make that change right now.

ppinchuk:
Hey, that's a great question. Yes, I ran a pipeline that included the following steps: 
"generation" -> "econ" -> "collect" -> "multi-year" -> "supply-curve-aggregation" -> "rep-profiles" .
Obviously these are not all the possible modules but I caught some issues that applied to all clis.

I will work on those example configs!

grantbuster:
Okay great, that's fine if there are a few growing pains but i just wanted to make sure these changes wouldn't be deadonarrival. 

Update version? Should note in the version logs that these will be breaking change.s

grantbuster (dismissed):

ppinchuk:
Yes, good call. I am assuming a minor version number?
By version logs do you mean the release?

PR #353, opened 2022-03-08
Bespoke

grantbuster:


reV/bespoke/gradient_free.py
```diff
@@ -0,0 +1,247 @@
...
+        mutation_rate : float, optional
+            The probability of a single bit mutating during the mutation phase
+            of the genetic algorithm.
+        tol : float, optional
+            The absolute tolerance to determine convergence.
+        convergence_iters : int, optional
+            The number of generations to determine convergence.
+        max_time : float
+            The maximium time (in seconds) to run the genetic algorithm.
+        """
```
grantbuster, line 50: Very detailed docstrings, excellent!

reV/bespoke/gradient_free.py
```diff
@@ -0,0 +1,247 @@
...
+
+    def optimize_ga(self):
+        """run the genetic algorithm"""
+
+        self.initialize_design_variables()
+        self.initialize_bits()
+        if self.population_size % 2 == 1:
+            self.population_size += 1
+        self.initialize_population()
+        self.initialize_fitness()
```
grantbuster, line 186: any reason these init methods shouldnt be called in the `__init__`? Is `optimize_ga()` intended to be called frequently? I wouldnt think so unless you were just running a bunch of random trials, but in that case i would think initializing many GA objects would be the intended pattern.

reV/bespoke/gradient_free.py
```diff
@@ -0,0 +1,247 @@
...
+        converged = False
+        ngens = 1
+        generation = 1
+        difference = self.tol * 10000.0
+        self.solution_history = np.zeros(self.max_generation + 1)
+        self.solution_history[0] = np.min(self.parent_fitness)
+
+        run_time = 0.0
+        start_time = time.time()
+        while converged is False and ngens < self.max_generation and \
```
grantbuster, line 197: Do you want to add any log statements to this module? See other reV modules for examples of how to set up a logger object for a module. Only add debug level logs because this module is called in parallel and info level would get crazy.

reV/bespoke/place_turbines.py
```diff
@@ -0,0 +1,234 @@
...
+                all SAM variables can be accessed
+                - cost: the annual cost of the wind plant (from cost_function)
+        cost_function : str
+            The cost function as a string, should return the annual cost
+            of the wind farm. Variables available are:
+                - n_turbines: the number of turbines
+                - system_capacity: wind plant capacity
+                - aep: annual energy production
+                - self.wind_plant: the SAM wind plant object, through which
+                all SAM variables can be accessed
```
grantbuster, line 45: Great docstrings. Clever that self.wind_plant is accessible to access any other SAM variables.

reV/bespoke/place_turbines.py
```diff
@@ -0,0 +1,234 @@
...
+        bits = np.ones(nlocs, dtype=int)
+        bounds = np.zeros((nlocs, 2), dtype=int)
+        bounds[:, 1] = 2
+        variable_type = np.array([])
+        for _ in range(nlocs):
+            variable_type = np.append(variable_type, "int")
+        ga = GeneticAlgorithm(bits, bounds, variable_type,
+                              self.optimization_objective)
+
+        ga.max_generation = 10000
```
grantbuster, line 190: This seems really sloppy. Why are these variables not set in the initialization function call `ga = GeneticAlgorithm(...)`?

reV/bespoke/place_turbines.py
```diff
@@ -0,0 +1,234 @@
...
+            self.capacity_density = 0.0
+
+        self.wind_plant["wind_farm_xCoordinates"] = \
+            self.x_locations[optimized_design_variables]
+        self.wind_plant["wind_farm_yCoordinates"] = \
+            self.y_locations[optimized_design_variables]
+        self.wind_plant["system_capacity"] = self.capacity
+        self.wind_plant.assign_inputs()
+        self.wind_plant.execute()
+        self.aep = self.wind_plant.annual_energy()
```
grantbuster, line 222: I would recommend that all variables from `self.turbine_x` on line 205 to `self.aep` on line 222 should actually be `@property` objects that retrieve their values on-the-fly instead of being assigned in this function call. Dynamically retrieved properties will ensure that these values are always current and are always self-consistent. It also allows you to add a small docstring describing each attribute which is really nice for documentation.

reV/bespoke/gradient_free.py
pjstanle: Agreed. I've moved these init methods to __init__

reV/bespoke/place_turbines.py
pjstanle: Agreed, this has been fixed. This actually threw an error when I updated '__init__' in the optimizer because it was trying to initialize a population size of zero

reV/bespoke/place_turbines.py
pjstanle: We'll need to talk about this, all/most of the variables are meaningless before running the optimization and are assigned with the optimal values of the optimization.

reV/bespoke/gradient_free.py
pjstanle: Yeah this is a good idea. I'll address this early next week

reV/bespoke/place_turbines.py
pjstanle: All of these variables are now accessed through attributes, and return None if they have not yet been assigned

PR #359, opened 2022-04-21
Bespoke upgrades and bug fixes

ppinchuk:
Addressed the following items:
- GA Algorithm now includes a fully packed solution as an initial parent
- GA Algorithm initialization `kwargs` can now be specified in a config file
  - Removed redundant `ga_time` argument, as it can now be specified using `ga_kwargs = {"max_time": time_in_seconds}`
- Fixed a bug that caused some outputs to be set to all 0 when written to file.
- Added a few more logging statements for to GA initialization

### DIFF
--- a/reV/bespoke/gradient_free.py
+++ b/reV/bespoke/gradient_free.py
@@ -52,15 +52,15 @@ class GeneticAlgorithm():
             The maximum time (in seconds) to run the genetic algorithm.
         """
 
-        logger.info('Initializing GeneticAlgorithm...')
-        logger.info('Minimum convergence iterations: {}'
-                    .format(convergence_iters))
-        logger.info('Max iterations (generations): {}'.format(max_generation))
-        logger.info('Population size: {}'.format(population_size))
-        logger.info('Crossover rate: {}'.format(crossover_rate))
-        logger.info('Mutation rate: {}'.format(mutation_rate))
-        logger.info('Convergence tolerance: {}'.format(tol))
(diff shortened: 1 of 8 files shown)


reV/bespoke/gradient_free.py
```diff
@@ -49,9 +49,19 @@ def __init__(self, bits, bounds, variable_type, objective_function,
...
 
+        logger.info('Initializing GeneticAlgorithm...')
+        logger.info('Minimum convergence iterations: {}'
+                    .format(convergence_iters))
+        logger.info('Max iterations (generations): {}'.format(max_generation))
+        logger.info('Population size: {}'.format(population_size))
+        logger.info('Crossover rate: {}'.format(crossover_rate))
+        logger.info('Mutation rate: {}'.format(mutation_rate))
+        logger.info('Convergence tolerance: {}'.format(tol))
+        logger.info('Maximum runtime (in seconds): {}'.format(max_time))
```
grantbuster, line 63: Can you just verify that these dont get parallel logged when doing multi-node multi-core jobs? I try to keep the parallelized logging statements as debug and the serial/manager level as info.

grantbuster (changes requested):

reV/bespoke/gradient_free.py
ppinchuk: Oops, I think they do get parallel logged. I will keep that in mind moving forward. Let me make adjustments!

grantbuster (approved):

PR #364, opened 2022-05-03
Gb/collect pattern

grantbuster:


reV/handlers/cli_collect.py
```diff
@@ -257,10 +247,9 @@ def get_node_cmd(name, h5_file, h5_dir, project_points, dsets,
     """
     # make a cli arg string for direct() in this module
     args = ['-f {}'.format(SLURM.s(h5_file)),
-            '-d {}'.format(SLURM.s(h5_dir)),
+            '-cd {}'.format(SLURM.s(collect_pattern)),
```
ppinchuk, line 250:
```suggestion
            '-cp {}'.format(SLURM.s(collect_pattern)),
```

Minor typo, but it breaks the whole cli ¯\\ _(ツ) _/¯

reV/config/collection.py
```diff
@@ -119,24 +100,52 @@ def _parse_pipeline_prefixes(self):
...
+    def fn_out_names(self):
+        """Get a list of output filenames ordered to correspond to the list of
+        collection patterns.
 
-        Returns
-        -------
-        file_prefixes : list
-            list of file prefixes to collect.
+        This can also be set to PIPELINE or just not set at all if this is a
+        pipeline job and collect_patterns="PIPELINE"
```
ppinchuk, line 108: I imagine the most typical use case will be not setting `fn_out_names` at all and `collect_patterns="PIPELINE"`, like you mention in the docstring here. This is exactly what I tried first, but it fails... If you follow the logic below for this case, you will hit the line `fn_out_names = list(fn_out_names)`, which gives you a `TypeError: 'NoneType' object is not iterable`

ppinchuk (changes requested):
Ok, added 2/3 bugs I found. The final bug was the fact that the output files did not end in ".h5", but I never traced this one down. I suggest we fix these first two and then I can run another test to see if the missing ".h5" file ending is even an issue anymore

ppinchuk (approved):
Couldn't find anything to complain about lolol. 
Nice job!

PR #366, opened 2022-05-12
Gb/force include range

grantbuster:


reV/supply_curve/exclusions.py
```diff
@@ -77,7 +86,7 @@ def __init__(self, layer, inclusion_range=(None, None),
         self._mask_type = self._check_mask_type()
 
     def __repr__(self):
-        msg = ("{} for {} exclusion, of type {}"
+        msg = ('{} for "{}" exclusion, of type "{}"'
```
ppinchuk, line 89: one thing I like to do instead of `"{}"` is `{!r}`, since it automatically adds quotation marks to string inputs. This is fine though!

ppinchuk (dismissed):
LGTM. Comments are very much "take it or leave it"

reV/supply_curve/exclusions.py
grantbuster: Yeah but not for numbers though so i kind of like the explicit quotation marks. I mean I think either is fine really, im going to leave it out of stubbornness :)

reV/supply_curve/exclusions.py
```diff
@@ -20,8 +20,11 @@ class LayerMask:
     Class to convert exclusion layer to inclusion layer mask
     """
     def __init__(self, layer, inclusion_range=(None, None),
-                 exclude_values=None, include_values=None,
-                 inclusion_weights=None, force_include_values=None,
+                 exclude_values=None,
```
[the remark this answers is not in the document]
grantbuster: yeahhhh thats a good idea. I'll do this.... tomorrow though haha

reV/supply_curve/exclusions.py
ppinchuk: No objections my friend!

reV/supply_curve/exclusions.py
```diff
@@ -247,19 +321,45 @@ def _check_mask_type(self):
...
+        ----------
+        data : ndarray
+            Exclusions data to create mask from
+
+        Returns
+        -------
+        mask : ndarray
+            Boolean mask of which values to include (True is include).
+        """
+        mask = True
```
ppinchuk, line 348: So I did not know that you could index into a `np.array` using just `True`, but I just tried it and it indeed works. However, it seems that doing so changes the shape of the array:

```
>>> import numpy as np
>>> a = np.arange(5)
>>> a
array([0, 1, 2, 3, 4])
>>> a.shape
(5,)
>>> a[True]
array([[0, 1, 2, 3, 4]])
>>> a[True].shape
(1, 5)
```

Do you think this would be an issue? Would it be worthwhile to add a test somewhere that has `exclude_range` set to something that does not intersect with the data? I'm not familiar enough with the rest of the exclusion code to know if this would make a difference

ppinchuk (changes requested):
I had just one more quick question, and if it's irrelevant, feel free to disregard and merge 😃

reV/supply_curve/exclusions.py
grantbuster: oh weirdddd. I just changed the init value of the mask to an np.fill. that should fix it.

ppinchuk: Fair enough!

ppinchuk (approved):

ppinchuk:
Excellent! This is now super flexible for anyone trying to set up exclusions, which is great because that job is difficult enough as it is!

PR #367, opened 2022-05-12
Pp/bespoke wake loss multiplier

ppinchuk:
Added option to scale wake losses **during the optimization step only**.

### DIFF
--- a/reV/bespoke/bespoke.py
+++ b/reV/bespoke/bespoke.py
@@ -50,7 +50,7 @@ class BespokeSinglePlant:
                  objective_function, capital_cost_function,
                  fixed_operating_cost_function,
                  variable_operating_cost_function,
-                 min_spacing='5x', ga_kwargs=None,
+                 min_spacing='5x', wake_loss_multiplier=1, ga_kwargs=None,
                  output_request=('system_capacity', 'cf_mean'),
                  ws_bins=(0.0, 20.0, 5.0), wd_bins=(0.0, 360.0, 45.0),
                  excl_dict=None, inclusion_mask=None, data_layers=None,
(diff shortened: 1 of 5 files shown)


grantbuster (approved):
This looks good but why wouldnt we apply the loss multiplier in the multiyear timeseries analysis if we apply it in the optimization? If @ppinchuk and the team have already talked about this, feel free to go ahead and merge.

ppinchuk:
The multiplier (as it's implemented) can only be applied on an annual scale, meaning we wouldn't be able to scale the wake loss contribution on an hourly level. This is the main reason it's left out of the final optimization results as well as any multiyear timeseries.

I believe (and please very much correct me if I'm wrong here), but the only way to pull out hourly wake losses is to do 2 reV runs... one for lossless generation and one for generation with wake losses. The two can then be compared to extract the wake loss profile and the use to scale it.

grantbuster:
Well pysam appears to output a [wake losses percent](https://nrel-pysam.readthedocs.io/en/master/modules/Windpower.html#PySAM.Windpower.Windpower.Outputs.wake_losses) variable... Would we want to use that to re-scale the capacity factors? The big problem with doing something like that is if the user requests something uncommon like `monthly_energy` that we haven't scaled... So I guess this depends on whether we want the wake loss multiplier to be applied just for the optimization or whether it would be a really nice thing for the output CF data as well.

ppinchuk:
Yes, that percentage value is for annual energy and it is what we use in the current implementation. I suppose *could* apply it as a haircut energy loss across the time series, but then you wouldn't get the hourly effect of wakes.

grantbuster:
Ah annual value. Got it. Let's merge this in and if we get the demand we can add in timeseries loss multiplier later.

PR #368, opened 2022-05-13
Pp/multi year losses seed

ppinchuk:
Generation cli `from_config` now seeds the outage losses so that a unique set of outages is generated for each year when running multi-year reV generation

### DIFF
--- a/reV/SAM/generation.py
+++ b/reV/SAM/generation.py
@@ -96,7 +96,7 @@ class AbstractSamGeneration(RevPySam, ScheduledLossesMixin, ABC):
             self.check_resource_data(resource)
             self.set_resource_data(resource, meta)
 
-        self.add_scheduled_losses()
+        self.add_scheduled_losses(resource)
 
     @classmethod
     def _get_res(cls, res_df, output_request):
--- a/reV/SAM/generation.py
+++ b/reV/SAM/generation.py
@@ -1220,6 +1220,7 @@ class WindPower(AbstractSamWind):
         return DefaultWindPower.default()
 
 
+# pylint: disable=too-many-ancestors
 class WindPowerPD(WindPower):
     """WindPower analysis with wind speed/direction joint probabilty
     distrubtion input"""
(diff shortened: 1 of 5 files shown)


grantbuster (changes requested):
Why not just add the year from the sam resource df to the seed value in `ScheduledLossesMixin`? This might be preferable to reduce CLI arguments (already too many haha).

ppinchuk:
Yea that's a great point haha. For some (strange and unknown) reason, I was trying to avoid having to pass `resource` to the `ScheduledLossesMixin`. In hindsight, this solution makes so much more sense than modifying the entire cli. Whoops! Thanks for keeping me honest here LOL! Will make that change...

ppinchuk:
The resource `DataFrame` is never stored as an attribute, so I unfortunately cannot access that. The `DataFrame` *is* parsed, but the data are stored under different keys (e.g. [solar_resource_data](https://github.com/NREL/reV/blob/239283364cd258460184f425d03c427b2f0045da/reV/SAM/generation.py#L591), [wind_resource_data](https://github.com/NREL/reV/blob/239283364cd258460184f425d03c427b2f0045da/reV/SAM/generation.py#L1209), etc.). Since the mixin class happens at the `AbstractSamGeneration` point (and I think it should stay that way), I would have to figure out what all of these keys are and check for them individually. Then, if we ever add new generation modules(e.g. geothermal), we would have to maintain the `ScheduledLossesMixin` class as well :/

The other thorn in all of this is that for some generation modules, [the resource data is not stored in `self` at all](https://github.com/NREL/reV/blob/239283364cd258460184f425d03c427b2f0045da/reV/SAM/generation.py#L1013).

grantbuster:
okay, your call on how you want to pass the info in! I'd just prefer minimizing argument passing especially on the cli

grantbuster (approved):

ppinchuk:
Yea I prefer less arguments as well, but I think in this case, this is the simplest solution - and with the current implementation the CLI is unchanged!

Issue #336, opened 2021-10-27
LCOT for SC Points that exceed 1500MW


MRossol:
**Bug Description**
The current hack for handling SC Points that exceed 1500MW is flawed. We currently connect those points to 500kV xmission features without changing the trans cap cost. When the trans cap cost is normalized by sc point capacity this produced abnormally low LCOTs.

**Full Traceback**
See e-mail thread with Patrick Brown and Trieu

**Expected behavior**
The original solution devised by Owen is the "correct" approach and would work as follows:
For SC points that > max tie-line capacity, the tie-line cost needs to be scaled by an integer multiplier equal to the # of tie-lines needed for the desired capacity and the trans_cap_cost recomputed. 
The multiplier would be computed as follows: `multiplier =  capacity // tie-line capacity + 1`

Typically the tie-line capacity would be 1500MW, but we should make this work for cases in which we don't pass in all the LC xmission classes

If we really want to be exact about this we should also consider re-computing the connection costs, namely the transformer cost...

**Additional context**
We pitched the "correct" solution to ReEDS and they suggested the hack (face palm emoji)

**Charge code**
ReV

grantbuster:
okay, path forward: do what michael suggested with the Nx transmission lines for SC points that exceed the max capacity, but then also when running reV for ReEDS maybe only feed in least cost transmission tables up to the 400MW max capacity lines so that it mimics Nx 400MW solar plants connected in parallel instead of single 4GW solar plants which are exceptional.

grantbuster:
An interesting effect of this is that an SC point could be connected to transmission via small capacity parallel lines even if a larger capacity line is available if the small parallel lines are more cost effective. Is this realistic? Seems reasonable?

grantbuster:
okay, seems reasonable.

grantbuster:
fixed in commit 4b83f1cef950b2ac55fe0c3c461a5ec86dffb8e6 and PR #370

PR #370, opened 2022-05-19
Gb/parallel spur lines

grantbuster:


grantbuster:
this was supposed to include commit https://github.com/NREL/reV/commit/4b83f1cef950b2ac55fe0c3c461a5ec86dffb8e6

In reference to issue #336

reV/supply_curve/supply_curve.py
```diff
@@ -251,25 +254,26 @@ def _map_trans_capacity(trans_sc_table):
         nx = trans_sc_table['capacity'] / trans_sc_table['max_cap']
         nx = np.ceil(nx).astype(int)
         trans_sc_table['n_parallel_trans'] = nx
-        mask = nx > 1
 
-        tie_line_cost = trans_sc_table.loc[mask, 'tie_line_cost'] * nx[mask]
+        if (nx > 1).any():
+            mask = nx > 1
```
ppinchuk, line 259: I am a little surprised a mask is needed here since, presumably, `nx[~mask] == 1` (unless capacity = 0, in which case `nx == 0` is still a valid multiplication to apply below).

I will assume there is a subtlety that I am missing here, but I figured I may as well point it out to you

ppinchuk (approved):
Can't honestly say I 100% understand everything that is happening, but I don't see any glaring issues

reV/supply_curve/supply_curve.py
grantbuster: No youre right, I think i found an annoying case where the input data was bad (maybe NaN's, i forget) and so i just wanted to be extra careful and not manipulate any data that didnt need manipulating.

ppinchuk: Ohhhhh, that makes sense. Ok so in that case, it's good to be extra careful!

PR #371, opened 2022-05-23
reV maintenance patch

ppinchuk:
Added many (small) unrelated changes that have piled up over time.

This PR looks really daunting, but it's mostly just documentation updates. I will try to summarize the code updates below (generally in the order of the changed files) for your convenience:
 - General documentation updates (typo fixes, updates to old links, and added missing modules).
 - Updated bespoke module to be compatible with losses (and vice versa, see below).
 - AEP explicitly set to 0 if no turbines are placed by bespoke.
 - Config files can now specify relative paths (i.e. `../project_points.csv`)
 - **Slightly** relaxed the string comparison of `crs` by splitting on empty spaces and doing set comparison (so that `"lat=0, lon=1"` is now equivalent to `"lon=1, lat=0"`). 
 - Fixed `dtype` bug in `reV/handlers/outputs.py`
 - Updated losses module to be compatible with bespoke (and vice versa, see above).
 - Fixed empty masks throwing an error in `reV/supply_curve/supply_curve.py`
 - Added tests for bespoke/losses linkage


grantbuster (approved):
Honestly looks good to me!

PR #372, opened 2022-05-24
Tw/turbine layout

WilliamsTravis:
This shifts the origin of the the relative x and y coordinates for the turbines from the bottom left of the bounding box around the available area (safe_polygons) to the bottom left of the full supply curve extent. This is important to be able to intuit the geographic locations of the individual turbine layouts using the relative coordinates included in the output meta.

### DIFF
--- a/reV/bespoke/place_turbines.py
+++ b/reV/bespoke/place_turbines.py
@@ -3,7 +3,6 @@
 place turbines for bespoke wind plants
 """
 import numpy as np
-import shapely.affinity
 
 from shapely.geometry import Polygon, MultiPolygon
 
--- a/reV/bespoke/place_turbines.py
+++ b/reV/bespoke/place_turbines.py
@@ -3,8 +3,9 @@
 place turbines for bespoke wind plants
 """
 import numpy as np
-from shapely.geometry import Polygon, MultiPolygon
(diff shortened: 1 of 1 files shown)


grantbuster:
@WilliamsTravis are you making the coordinates relative to the bottom left of the full exclusion raster like relative to some point offshore by san diego or relative to the full SC point?

If it's the former, I'd rather not feed SAM the coordinates relative to all of conus if possible, even if that doesnt break things thats definitely not the intended input. An easy alternative if its just a convenience thing is you can easily calculate the offset from the bottom left of the exclusions file using the sc_row_ind and sc_col_ind * 90m.

WilliamsTravis:
They are now relative to the corner of each supply curve point, so the latter. PJ and I just sat to down to find out why some layouts in reView were rendering correctly while other weren't. This was happening because the origin of the "turbine_x_coords" and "turbine_y_coords" were relative to a bounding box around the included area after exclusions. So, for example, in the case where the available area was completely contained in the upper right quadrant of a cell, the origin would in the center and turbine coordinates relative to that.

grantbuster:
Ah gotcha, sounds good!

grantbuster (approved):

PR #375, opened 2022-06-21
subclassed outputs from rex. incremented rex version requirement.

bnb32:


grantbuster (changes requested):

reV/handlers/outputs.py
```diff
@@ -210,772 +141,25 @@ def package(self):
         return self.h5.attrs['package']
```
grantbuster, line 141: Why do we need this package property copied in the reV handler?

bnb32: we don't lol. removed.

grantbuster (approved):

PR #385, opened 2022-10-19
Gb/min area pixel area

grantbuster:


tests/test_gen_pv.py
```diff
@@ -247,8 +249,6 @@ def test_pv_name_error():
 
     # run reV 2.0 generation
     with pytest.raises(KeyError) as record:
-        pp = ProjectPoints(rev2_points, sam_files, 'pv',
```
ppinchuk, line 250: So weird that these lines were left in the test with seemingly no effect. Good catch on cleaning those up

ppinchuk (commented):
Looks clean enough - though >90% of the changes had nothing to do with the min pixel area ;)
Strange that the tests for Python 3.8+ are failing but not for 3.7. Any idea what's going on?

ppinchuk (approved):
Nice job!

grantbuster:
Yeah there was a weird pandas.DataFrame.at[] indexing operation that some versions didnt like, super weird but easy fix: https://github.com/NREL/reV/pull/385/commits/9d2dcbb2ddf6c907db5c018c17be83c287f769b7

PR #390, opened 2022-12-09
Bnb/collection fix

bnb32:


grantbuster (dismissed):
looks good! glad we fixed this. Probably going to start running a lot of gid_map projects as we get more resource datasets.

PR #391, opened 2022-12-18
Bespoke bug fixes

ppinchuk:
Fix for several bugs:
- Bespoke run no longer crashes if all input points are fully excluded. 
- Previous `exclude_range` logic was looking for values that are both less than the min value and larger than the max value simultaneously. This was causing all layers using `exclude_range` to be fully excluded every time. 
- Bespoke no longer crashes if data layers do not specify `fpath`. This now matches reV-gen behavior, and was accomplished by moving the file opening logic into the points class

All logic changes include tests.

### DIFF
--- a/reV/bespoke/bespoke.py
+++ b/reV/bespoke/bespoke.py
@@ -1312,7 +1312,10 @@ class BespokeWindPlants(AbstractAggregation):
             Full filepath to an output .h5 file to save Bespoke data to. The
             parent directories will be created if they do not already exist.
         """
-
+        if not self.completed_gids:
+            logger.info('No output data found! It is likely that all '
+                        'requested points are excluded.')
+            return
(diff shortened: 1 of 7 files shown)


reV/bespoke/bespoke.py
```diff
@@ -1312,7 +1312,10 @@ def save_outputs(self, out_fpath):
             Full filepath to an output .h5 file to save Bespoke data to. The
             parent directories will be created if they do not already exist.
         """
-
+        if not self.completed_gids:
```
grantbuster, line 1315: Probably worth a warning

grantbuster (approved):

reV/bespoke/bespoke.py
ppinchuk: Great point, updated to a warning here

PR #392, opened 2022-12-23
Bespoke site data to sys inputs

ppinchuk:
Added ability to input site-specific SAM technology inputs via the project points csv input in Bespoke. 
Also added support for capital cost, fixed operating cost, and variable operating cost multiplier inputs, also via project points.

Also simplified some of the code (no logic changes) in the `PlaceTurbines` class. 

No new tests were necessary - changes were small enough to fit within the existing test infrastructure.

### DIFF
--- a/reV/bespoke/bespoke.py
+++ b/reV/bespoke/bespoke.py
@@ -1217,6 +1217,34 @@ class BespokeWindPlants(AbstractAggregation):
             meta = meta[0]
         return meta
 
+    def sam_sys_inputs_with_site_data(self, gid):
+        """Update the sam_sys_inputs with site data for the given GID.
+
+        Site data is extracted from the project points DataFrame. Every
+        column in the project DataFrame becomes a key in the site_data
+        output dictionary.
+
+        Parameters
+        ----------
+        gid : int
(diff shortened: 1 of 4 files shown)


grantbuster (dismissed):

reV/bespoke/bespoke.py
```diff
@@ -1217,6 +1218,34 @@ def meta(self):
...
+            DataFrame.
+        """
+
+        gid_idx = self._project_points.index(gid)
+        site_data = self._project_points.df.iloc[gid_idx]
+
+        site_sys_inputs = self._project_points[gid][1]
+        site_sys_inputs.update({k: v for k, v in site_data.to_dict().items()
+                                if not isinstance(v, float)
+                                or not np.isnan(v)})
```
grantbuster, line 1246: This logic is confusing but its basically don't set if float and NaN right? I always prefer writing it how you would say it so like `if not (float and NaN)`

ppinchuk: Fair enough, I'll update that

grantbuster (approved):

PR #394, opened 2023-01-12
Bnb/dev

bnb32:
Added flag `write_mapped_gids` flag so the project point gids will be written to output meta instead of the resource gids.

### DIFF
--- a/reV/config/sam_analysis_configs.py
+++ b/reV/config/sam_analysis_configs.py
@@ -217,6 +217,17 @@ class GenConfig(SAMAnalysisConfig):
         return self.get('gid_map', None)
 
     @property
+    def write_mapped_gids(self):
+        """Option to write mapped gids to meta data gids instead of the gids
+        from the resource data.
+
+        Returns
+        -------
+        bool
+        """
+        return self.get('write_mapped_gids', False)
+
+    @property
     def resource_file(self):
         """
         get base resource_file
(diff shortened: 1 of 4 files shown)


ppinchuk (approved):
Code-wise it LGTM. Before merging, we should make sure there is no other way to achieve this behavior. Maybe @grantbuster can chime in?

grantbuster (changes requested):

grantbuster:
> Code-wise it LGTM. Before merging, we should make sure there is no other way to achieve this behavior. Maybe @grantbuster can chime in?

I think this is fine, @ppinchuk do you have a concern that i'm not seeing? The gid_map is kind of a hack and I think it makes sense that we do the replacement in the meta data property where the gid_map takes place.

grantbuster (approved):

ppinchuk:
> > Code-wise it LGTM. Before merging, we should make sure there is no other way to achieve this behavior. Maybe @grantbuster can chime in?
> 
> I think this is fine, @ppinchuk do you have a concern that i'm not seeing? The gid_map is kind of a hack and I think it makes sense that we do the replacement in the meta data property where the gid_map takes place.

No concerns, I think this is good. Just wanted to make sure to get your eyes on it too. I've only used gid_map once I think, so you'd be more familiar with it than I am

Issue #374, opened 2022-06-06
POI lon, lat in suppl curve table


alopez04:
**Why this feature is necessary:**
We'd like to add the latitude and longitude of the point of interconnection for wind (LBW & OSW) and solar supply curve points. Here's the scenario, a reV supply curve point is in offshore waters and is assigned a county in New Jersey, but the supply curve point has a point of interconnection in New York City. It matters a lot more for downstream modeling where the energy is going, not where the farm sits relative to state/county borders. Same is true for LBW and solar.  

**A possible solution is:**
Additional two columns in the resulting supply curve table that specify the latitude and longitude of the POI used. 

**I have considered the following alternatives:**
None. This is the correct approach ;)

**Additional context**
![image](https://user-images.githubusercontent.com/9019826/172263574-9f0ca44b-573f-480f-98c7-c1ca198cb729.png)

**Charge code**
WEPG.10348.08.01.06

**Urgency / Timeframe**
Not urgent / July 2022

ppinchuk:
As of #396, reV will pass through the `poi_lat` and `poi_lon` columns, if they are already given in the transmission tables. Legacy transmission tables will not have those columns, but any transmission tables created [after this reVX update](https://github.com/NREL/reVX/pull/172) should have those columns (including the reinforcement costs tables). I think with that, we can close this issue, unless you have any other thoughts on this @grantbuster

grantbuster:
Sounds good!

PR #396, opened 2023-01-27
Reinforcement Cost Support

ppinchuk:
This PR is for compatibility with the [reVX Reinforcement Cost addition.](https://github.com/NREL/reVX/pull/172)

Mostly comprises of minor logic to support sorting on a transmission table that includes reinforcement costs. 

Also added logic for to passthrough poi lat/lon if they are given in the transmission table (see #374).
Along the same lines, added two new ReEDS passthrough columns that will be added to bespoke in a PR coming in the very near future.

### DIFF
--- a/reV/supply_curve/supply_curve.py
+++ b/reV/supply_curve/supply_curve.py
@@ -735,6 +735,8 @@ class SupplyCurve:
             self._trans_table['trans_cap_cost_per_mw'] = cost
 
         if 'reinforcement_cost' in self._trans_table:
+            logger.info("'reinforcement_cost' column found in transmission "
+                        "table. Adding reinforcement costs to total LCOE.")
             cf_mean_arr = self._trans_table['mean_cf'].values
             lcot = (cost * fcr) / (cf_mean_arr * 8760)
             lcoe = lcot + self._trans_table['mean_lcoe']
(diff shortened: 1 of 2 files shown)


reV/supply_curve/supply_curve.py
```diff
@@ -327,9 +327,8 @@ def _check_sub_trans_lines(cls, features):
         if not any(mask):
             return []
 
-        line_gids = \
-            features.loc[mask,
-                         'trans_line_gids'].apply(cls._parse_trans_line_gids)
+        line_gids = (features.loc[mask, 'trans_line_gids']
+                     .apply(cls._parse_trans_line_gids))
```
grantbuster, line 331: way better formatting haha

reV/supply_curve/supply_curve.py
```diff
@@ -735,6 +734,18 @@ def compute_total_lcoe(self, fcr, transmission_costs=None,
             cost /= self._trans_table['capacity']  # $/MW
             self._trans_table['trans_cap_cost_per_mw'] = cost
 
+        if 'reinforcement_cost' in self._trans_table:
```
grantbuster, line 737: maybe add an info logging statement if this is found? I always find that i put these conditionals in here, then it doesn't work, and i spend a ton of time debugging when i just made a typo in the string or the table header. It's also nice to know that this bit of code is running.

reV/supply_curve/supply_curve.py
```diff
@@ -966,10 +981,34 @@ def _check_feature_capacity(self, avail_cap_frac=1):
...
 
+    def _adjust_sort_inputs(self, columns, sort_on, consider_friction):
+        """Add extra output columns and set `sort_on` value, if needed. """
+        # These are essentially should-be-defaults that are not
+        # backwards-compatible, so have to explicitly check for them
+        extra_cols = {'poi_lat', 'poi_lon', 'reinforcement_poi_lat',
+                      'reinforcement_poi_lon', 'eos_mult', 'reg_mult'}
+
+        if consider_friction and 'total_lcoe_friction' in self._trans_table:
+            extra_cols.add('total_lcoe_friction')
```
grantbuster, line 992: why don't you just iterate through these extra cols? What about:
```
extra_cols = {all, the, options}
extra_cols = set(col for col in extra_cols if col in self._trans_table)

if 'reinforcement_cost' in self._trans_table:
    sort_on = sort_on or "lcoe_no_reinforcement"

File: tests/test_supply_curve_compute.py [hunk 4]
```diff
@@ -328,3 +330,154 @@ def test_multi_parallel_trans():
...
+        verify_trans_cap(sc_full, trans_tables)
+
+        trans_tables = []
+        for cap in [100, 200, 400, 1000]:
+            in_table = os.path.join(TESTDATADIR, 'trans_tables',
+                                    f'costs_RI_{cap}MW.csv')
+            in_table = pd.read_csv(in_table)
+            out_fp = os.path.join(td, f'costs_RI_{cap}MW.csv')
+            in_table["reinforcement_cost"] = 1e6
+            in_table["reinforcement_dist_km"] = 10
```
grantbuster, line 371: really like how weve been doing this lately - add tests that manipulate the existing files with simple checks instead of adding new baseline inputs/outputs

grantbuster (changes requested):

tests/test_supply_curve_compute.py
ppinchuk: Me too :) And the code spells out what is important about the data in the file, so you don't have to go digging thought the file manually to find out

reV/supply_curve/supply_curve.py
ppinchuk: Good point! Added below

reV/supply_curve/supply_curve.py
ppinchuk: Yea that's a great point. I've changed the code to match that style and split apart the unrelated functionality of determining the output columns and the `sort_on` column into separate functions

grantbuster (approved):

PR #397, opened 2023-02-03
New Columns

ppinchuk:
Added `eos_mult` and `reg_mult` columns to bespoke meta. The are computed directly from bespoke results and are required for ReEDS runs. Also deprecated `reinforcement_cost` in favor of `reinforcement_cost_per_mw` since [the former will no longer be part of new reinforcement tables](https://github.com/NREL/reVX/pull/174)

### DIFF
--- a/reV/bespoke/place_turbines.py
+++ b/reV/bespoke/place_turbines.py
@@ -329,6 +329,35 @@ class PlaceTurbines:
         self.initialize_packing()
         self.optimize(**kwargs)
 
+    def capital_cost_per_kw_at(self, capacity_mw):
+        """Capital cost function (per kW) evaluated for a given capacity.
+
+        The capacity will be adjusted to be an exact multiple of the
+        turbine rating in order to yield an integer number of
+        turbines.
+
+        Parameters
+        ----------
+        capacity_mw : float
+            The desired capacity (MW) to sample the cost curve at. Note
+            as mentioned above, the capacity will be adjusted to be an
+            exact multiple of the turbine rating in order to yield an
+            integer number of turbines. For best results, set this
+            value to be an integer multiple of the turbine rating.
+
+        Returns
+        -------
+        capital_cost : float
+            Capital cost (per kW) for the (adjusted) plant capacity.
+        """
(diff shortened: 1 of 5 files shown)


reV/bespoke/place_turbines.py
```diff
@@ -329,6 +329,35 @@ def place_turbines(self, **kwargs):
...
+            The desired capacity (MW) to sample the cost curve at. Note
+            as mentioned above, the capacity will be adjusted to be an
+            exact multiple of the turbine rating in order to yield an
+            integer number of turbines. For best results, set this
+            value to be an integer multiple of the turbine rating.
+
+        Returns
+        -------
+        capital_cost : float
+            Capital cost (per kW) for the (adjusted) plant capacity.
```
grantbuster, line 351: $ per kW yeah?

reV/bespoke/place_turbines.py
```diff
@@ -329,6 +329,35 @@ def place_turbines(self, **kwargs):
         self.initialize_packing()
         self.optimize(**kwargs)
 
+    def capital_cost_per_kw_at(self, capacity_mw):
```
grantbuster, line 332: What's up with the `_at` suffix? "At the input capacity"? Kind of funky naming IMO but fine if you like it.

grantbuster (approved):

reV/bespoke/place_turbines.py
ppinchuk: Yea I guess the thought was to make it read like English: `capital_cost_per_kw_at(200)`.
Happy to change it to  `capital_cost_per_kw`, I don't have a strong preference here

reV/bespoke/place_turbines.py
ppinchuk: Ooo yes, I should clarify

reV/bespoke/place_turbines.py
grantbuster: Yeah gotcha. This is an interesting one haha I haven't thought to do this. I could go either way!

PR #398, opened 2023-02-08
Gb/multi res

grantbuster:
Multi-resolution resource handling feature. Tests will fail until we merge the rex branch, but I want this to be tested on Eagle with dev installs before we do.

PR #399, opened 2023-02-09
reV Geothermal Module V1

ppinchuk:
First version of reV Geothermal Module.

This update allows reV to run the SAM-GETEM module in the same fashion as wind/solar. Only generation is impacted - reV functions just like the other technologies for the aggregation and supply curve modules. Representative profiles may need to be adjusted to support the geothermal module in the future.

Other tweaks may come in a future patch if the results indicate the need.

This update also includes an upgrade to PySAM 4.


ppinchuk:
Will convert from draft to actual PR after #398 is merged (don't want to conflict with versioning)

reV/SAM/generation.py
```diff
@@ -1149,6 +1161,221 @@ def default():
...
+                   'setting to `False` (0)!')
+            logger.warning(msg)
+            warn(msg)
+            self.sam_sys_inputs["ui_calculations_only"] = 0
+        super().assign_inputs()
+
+    def set_resource_data(self, resource, meta):
+        """Generate the weather file and set the path as an input.
+
+        The Geothermal PySAM model requires a data file, not raw data.
```
grantbuster, line 1207: ugh another one of these huh? annoying that this is still a thing. Nice work using the other similar frameworks though.

reV/SAM/generation.py
```diff
@@ -1149,6 +1161,221 @@ def default():
...
+
+    def _set_resource_potential_to_match_gross_output(self):
+        """Set the resource potential input to match the gross generation.
+
+        As of 12/20/2022, the resource potential input is only used to
+        calculate the number of well replacements during the lifetime
+        of a geothermal plant. It was decided that reV would not model
+        well replacements, and therefore it is fine to set the resource
+        potential to match (or be just above) the gross potential so
+        that SAM does not throw any errors.
```
grantbuster, line 1282: good documentation and implementation. Consider putting this blurb in the class docstring though. Hidden methods don't get an html docs thing in the gh pages.

reV/SAM/generation.py
```diff
@@ -1149,6 +1161,221 @@ def default():
...
+        return fname
+
+    def run_gen_and_econ(self):
+        """Run SAM generation and possibility follow-on econ analysis."""
+        try:
+            super().run_gen_and_econ()
+        except SAMExecutionError as e:
+            logger.error("Skipping site {}; received sam error: {}"
+                         .format(self._site, str(e)))
+            self.outputs = {}
```
grantbuster, line 1376: Would we want default outputs? e.g. capacity factor = 0? What outputs get written if the sim fails?

tests/test_gen_geothermal.py
```diff
@@ -0,0 +1,147 @@
...
+            if len(test.shape) == 2:
+                test = np.mean(test, axis=0)
+
+            msg = ('{} outputs do not match baseline value! Values differ '
+                   'at most by: {}'
+                   .format(dset, np.max(np.abs(truth - test))))
+            assert np.allclose(truth, test, rtol=RTOL, atol=ATOL), msg
+
+
+def test_gen_geothermal_temp_too_low():
```
grantbuster, line 80: Is this a negative test where the SAM simulation will fail? If not, can we add one of those?

grantbuster (changes requested):
Nice work! Way less code than i had feared haha.

reV/SAM/generation.py
ppinchuk: Yea, don't even get me started lol. *They don't even use the weather data*.

reV/SAM/generation.py
ppinchuk: Yea good point, I will move that up

reV/SAM/generation.py
ppinchuk: Yea the reV `Gen` class initializes all outputs to 0, so those are the default outputs. If the sim fails, the default initialized value is *not* overwritten, so the outputs are all 0 for that point.

Happy to walk through this in finer detail during a short call if you'd like!

tests/test_gen_geothermal.py
ppinchuk: Yup, this is the negative test! PySAM throws an error for very low temps, so in this test I check to make sure the outputs are all set to 0 (the initialized default value - see above question) in the output generation file.

PR #401, opened 2023-02-20
Pp/skip pc losses

grantbuster:


reV/losses/power_curve.py
```diff
@@ -487,8 +536,28 @@ def add_power_curve_losses(self):
             return
 
         wind_resource, weights = self.wind_resource_from_input()
-        pc_losses = PowerCurveLosses(self.input_power_curve,
-                                     wind_resource, weights)
+        power_curve = self.input_power_curve
+
+        if (wind_resource <= power_curve.cutin_wind_speed).all():
```
ppinchuk, line 541: One weird case we are missing here is if there are wind speeds both above the cutoff speed and below the cutin speed. I know this is unlikely, but it would be easy to check for all three cases in one go: 

```
if not power_curve(wind_resource).any():
    msg = ("Input power curve does not generate power for any of "
           "the wind speeds at site {}. No power curve adjustments made!"
           .format(getattr(self, "site", "[unknown]"))
    logger.warning(msg)
    warnings.warn(msg, reVLossesWarning)
    return
```
Not sure what your appetite for making this change is though - I won't impose either way

tests/test_losses_power_curve.py
```diff
@@ -228,6 +229,39 @@ def get_item_patch(self, key):
...
+    PowerCurveLossesMixin.__getitem__ = get_item_patch
+
+    mixin = PowerCurveLossesMixin()
+    mixin.sam_sys_inputs = copy.deepcopy(sam_config)
+    mixin.sam_sys_inputs[PowerCurveLossesMixin.POWER_CURVE_CONFIG_KEY] = {
+        'target_losses_percent': 10,
+        'transformation': 'horizontal_translation'
+    }
+    mixin.sam_sys_inputs['wind_resource_data'] = {
+        'data': [(1, 10, 1_000, 0) for __ in BASIC_WIND_RES]
```
ppinchuk, line 253: Would be good to do a similar test for low wind speeds and/or a mix of the two, but if the PR is urgent, I won't impose

ppinchuk (approved):
Sorry you had to deal with these weird cases. This is what I get for relying on third-party packages to do what they promise to do eh? lol

Anyways, some optional suggestions below, but overall looks good.

reV/losses/power_curve.py
grantbuster: ohhh you're saying if the wind speed timeseries is always low but then has random gusts greater than the cutout? Sounds pretty unlikely haha i say we keep it as is so we get more meaningful logging.

tests/test_losses_power_curve.py
grantbuster: Yep agreed, will add!

ppinchuk (approved):

PR #402, opened 2023-02-22
Updated Geothermal Cost Inputs

ppinchuk:
Since reV dynamically sets the size of a geothermal plant, it should update the associated costs accordingly. This patch allows users to input `capital_cost_per_kw` and `fixed_operating_cost_per_kw` instead of the flat `capital_cost` and `fixed_operating_cost` values, respectively. If these inputs are detected, reV calculates the total `capital_cost` and `fixed_operating_cost` based on the plant size and automatically adds them to the SAM config on a per-site basis.

### DIFF
--- a/reV/SAM/generation.py
+++ b/reV/SAM/generation.py
@@ -1181,6 +1181,7 @@ class Geothermal(AbstractSamGenerationFromWeatherFile):
     MODULE = 'geothermal'
     PYSAM = PySamGeothermal
     PYSAM_WEATHER_TAG = "file_name"
+    _RESOURCE_POTENTIAL_MULT = 1.001
 
     @staticmethod
     def default():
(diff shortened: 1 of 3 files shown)


grantbuster (approved):
Makes sense and looks good. I'm wondering if there's any way to document some of these hidden options? We have several features/options baked into the SAM config and we've never figured out how to document them appropriately. Maybe just in the SAM generation class docstring?

ppinchuk:
Yea, probably putting this into the SAM class is a minimum. I will also link the `OPTIONS` dictionary to the `reV_run` function documentation to encourage users to look at the relevant SAM class docs, but there is no guarantee that will happen. I will continue to think about the best way to round these up and document them...

grantbuster (approved):
This is great. We should make sure we do this when we add more sam-config features. The only one i can think of right now is the `time_index_step` arg.

PR #403, opened 2023-03-03
Gb/bc

grantbuster:


reV/generation/generation.py
```diff
@@ -481,6 +501,61 @@ def _parse_nn_map(self):
...
+            resource bias correction table. This has columns: gid (can be index
+            name), adder, scalar. If both adder and scalar are present, the
+            wind or solar resource is corrected by (res*scalar)+adder. If
+            either adder or scalar is not present, scalar defaults to 1 and
+            adder to 0. Only windspeed or GHI+DNI are corrected depending on
+            the technology. GHI and DNI are corrected with the same correction
+            factors.
+        """
+
+        if isinstance(bias_correct, type(None)):
```
ppinchuk, line 531: Ooo, genuinely curious - why `type(None)` here? Never seen this before, is it more robust?

ppinchuk (approved):
Nice, great little feature for reV :)

reV/generation/generation.py
grantbuster: Haha great question, i think they're equivalent, i think i just used isinstance to match the syntax of the elif piece.

PR #405, opened 2023-03-09
PySam 4.1.0 Support

ppinchuk:
Mostly needed for reV geothermal - this fixes CF > 1 issue


requirements.txt
```diff
@@ -1,5 +1,5 @@
 NREL-NRWAL>=0.0.7
-NREL-PySAM~=4.0.0
+NREL-PySAM~=4.1.0
```
grantbuster, line 2: What is the tilda again? Thoughts on just pinning a single version with ==? Would ensure compatibility and reproducibility. The only downside i see is we wouldn't know when new sam versions release because our code wouldn't break (lol)

grantbuster (approved):

requirements.txt
ppinchuk: Ah shoot, this comment was hidden for some reason, so I didn't see it before merging. The tilda just means "get the most up to date minor version". So for example, if the SAM team releases 4.1.1, we would get that update (which technically should only include bug/security fixes), but if they then release 4.2.0, we would still only get 4.1.1. Essentially it's pinning to an API version but getting the most bug/security concern free version. 

That of course assumes the SAM team properly follows semantic versioning guidelines. 
I'm happy to pin it with == in the next PR

grantbuster: Ah no, this sounds perfect

PR #407, opened 2023-04-20
Gb/bespoke gen

grantbuster:


reV/bespoke/bespoke.py
```diff
@@ -277,6 +303,64 @@ def _parse_output_req(self):
             logger.error(msg)
             raise KeyError(msg)
 
+    def _parse_prior_run(self):
+        """Parse prior bespoke wind plant optimization run meta data and make
+        sure the SAM system inputs are set accordingly."""
+
+        # {meta_column: sam_sys_input_key}
+        required = {'capacity': 'system_capacity',
```
ppinchuk, line 311: Should we document these required columns in the docstring for `prior_meta` above?

reV/bespoke/bespoke.py
```diff
@@ -277,6 +303,64 @@ def _parse_output_req(self):
...
+                    'turbine_x_coords': 'wind_farm_xCoordinates',
+                    'turbine_y_coords': 'wind_farm_yCoordinates'}
+
+        if self._prior_meta:
+            missing = [k for k, _ in required.items() if k not in self.meta]
```
ppinchuk, line 316: Just curious, why not simply `missing = [k for k in required if k not in self.meta]` here?

ppinchuk:
I started a review but I think this is still WIP(?), so I will finish later

grantbuster:
@ppinchuk no it's ready for review. Just had some unexpected tests fail so had to fix some things. Tests look good now.

tests/test_bespoke.py
```diff
@@ -983,3 +913,229 @@ def test_bespoke_aep_is_zero_if_no_turbines_placed():
...
+        cols = ['n_gids', 'gid_counts', 'res_gids']
+        pd.testing.assert_frame_equal(meta1[cols], meta2[cols])
+
+        assert not np.allclose(data1['cf_mean-2013'], data2['cf_mean-2013'])
+        assert not np.allclose(data1['ws_mean'], data2['ws_mean'], atol=0.2)
+        assert np.allclose(ws.mean(), data2['ws_mean'], atol=0.01)
+
+
+if __name__ == '__main__':
+    output_request = ('system_capacity', 'cf_mean', 'cf_profile',
```
ppinchuk, line 1076: This is the bias correct test, right? Should it be in it's own test function or is this intentional?

ppinchuk (commented):
@grantbuster Nice work. A bit unfortunate that we have two parallel generation workflows  to maintain now, but I am glad we (you :P) are being diligent about keeping them in sync.

I noticed there are no CLI updates for these inputs, so they are not "available" in the general sense for runs yet. Are you not expecting to need this functionality for the next few months (while reV is transitioning to GAPs)?

reV/bespoke/bespoke.py
grantbuster: Oh you know what, i was doing in required.keys() but pylint got angry and i think i misinterpreted the message as "use items" instead of "iterate directly"

ppinchuk (approved):
LGTM, nice work!

PR #409, opened 2023-04-26
missed python version increment in install instructions

bnb32:


grantbuster (approved):
in the future you can also just edit README files directly in github on main.

bnb32:
> in the future you can also just edit README files directly in github on main.

I thought I could but it made me submit a PR. Unless I missed something.

PR #410, opened 2023-05-03
Bespoke Updates

ppinchuk:
Mostly a collection of improvements to address annoyances I ran into while running bespoke for ATB turbines.

Also added the option to pre-load resource data before running bespoke for all sites. I've gotten mixed results with this feature so far. Initial testing seemed very promising, but more recent tests paint a less definitive picture of improvement. In any case, the code has already been implemented, and it may not be a bad thing to leave it as a toggleable option for the user.

This is Part 1 of my planned improvements for bespoke - mostly focused on quality-of-life adjustments. I will be adding a few logic-focused features in Part 2, but no ETA for that yet.

### DIFF
--- a/reV/bespoke/bespoke.py
+++ b/reV/bespoke/bespoke.py
@@ -4,6 +4,7 @@ reV bespoke wind plant analysis tools
 """
 # TODO update docstring
 # TODO check on outputs
+import time
 import logging
 import copy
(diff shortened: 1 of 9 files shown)


ppinchuk:
Merging to move some work along. Happy to circle back to any of these changes if need be

PR #414, opened 2023-05-16
added better parsing for windspeed and winddirection variables in the…

grantbuster:
… bespokeoutput request


ppinchuk (approved):
Nice, more flexibility with outputs is always appreciated

PR #415, opened 2023-05-18
reV on GAPs

ppinchuk:
This update contains quite a few changes, mostly to the reV CLI. Primarily, all CLI functionality has been moved to [GAPs](https://github.com/NREL/gaps). This means a lot of the reV code having to do with the CLI has been deprecated. 

Most importantly, a big documentation overhaul has been included.

This update also includes fixes for a few outstanding issues.

Since GAPs is not available on the platform, `conda` support is dropped with this update.

`reVX` tests will fail due to breaking API changes. Will update reVX code to confirm to new changes once this PR is merged.


ppinchuk:
Ok @grantbuster, take a peek at the new [`Gen`](https://github.com/NREL/reV/blob/cac5a1876ee6a46e7e4674c026a53406f48bb777/reV/generation/generation.py#L52) class. In particular, note that  [`reV_run`](https://github.com/NREL/reV/blob/cac5a1876ee6a46e7e4674c026a53406f48bb777/reV/generation/generation.py#L620) is now an instance method instead of a class method. This allows the docstring for the parameters to be broken up between the instance method and the class init. All parameters only have to be documented in the one place where they are pulled into the object. Note the [new way of using the Gen class](https://github.com/NREL/reV/blob/cac5a1876ee6a46e7e4674c026a53406f48bb777/reV/generation/generation.py#L69-L97)

```
gen = Gen(tech, points, sam_configs, res_file, lr_res_file=None,
          output_request=('cf_mean',), site_data=None, curtailment=None,
          gid_map=None,  sites_per_worker=None,
          mem_util_lim=0.4, scale_outputs=True, write_mapped_gids=False,
          bias_correct=None)
gen.reV_run(max_workers=1, pool_size=(os.cpu_count() * 2), timeout=1800)
```

does resemble [the old class method implantation](https://github.com/NREL/reV/blob/be11388c049e002c02cb599f040f98e5619e022e/reV/generation/generation.py#L65-L91): 

```
gen = Gen.reV_run(tech, points, sam_configs, res_file, lr_res_file=None,
                  output_request=('cf_mean',), site_data=None, curtailment=None,
                  gid_map=None, max_workers=1, sites_per_worker=None,
                  pool_size=(os.cpu_count() * 2), timeout=1800,
                  points_range=None, out_fpath=None, mem_util_lim=0.4,
                  scale_outputs=True, write_mapped_gids=False,
                  bias_correct=None)
```

No docstrings have actually been updated yet, I have focused purely on the code logic for now.

Let me know if you see any major red flags or funky things you'd like me to address.

ppinchuk:
All great questions! 

1. I plan on removing all existing logic in `reV/generation/cli_gen.py` completely. There is still [some CLI-specific logic](https://github.com/NREL/reV/blob/84408db73bdc5d8052d1ae1391ede5f801070a44/reV/generation/generation.py#L700-L768) that I was thinking about moving to `reV/generation/cli_gen.py`, but I don't really have strong feelings about that. I would be happy to keep that in the generation file if you think that is most appropriate.
2. I do plan on removing _most_ of the config classes. I definitely plan on deprecating the generation config and other command configs. Some configs do seem to have a use though (e.g. Curtailment config). I was planning on re-evaluating their structure towards the end of the refactor to see if they should remain as-is or if there are ways to simplify them.
3. Yes, definitely. I like the `Gen.reV_run() -> Gen.run()` change a lot so I would be perfectly happy to do that.
4. Yes definitely. I will send you a local copy of them so that you can explore, but I would also be happy to jump on a call and discuss.

ppinchuk:
Noted on all of those. And yes, people can still totally run the commands locally (but they will need to use a config)

grantbuster:
> Noted on all of those. And yes, people can still totally run the commands locally (but they will need to use a config)

Yep got it. and can run from python if they want to. PERFECT.

PR #416, opened 2023-05-31
Solar AC capacity outputs

ppinchuk:
Add option to outputs AC capacity and capacity factor for solar.
Also allow transmission to be sized on a column other than "capacity" (e.g. "capacity_ac").


grantbuster (approved):
This looks good, but ReEDS uses rep profiles as input a lot. Do we need to modify that at all or can we just use a different `cf_dset` input?

grantbuster:
> Yea I looked through the rep-profiles code but I think all it uses are the GID values from the supply curve. We should be able to just change the `cf_dset` inputs, and that's exactly what I am doing right now for the solar standard scenario runs. Things seem to be working so far, but I can wait to merge this until I confirm that those outputs look good.

Excellent!

ppinchuk:
Rep profiles look good as far as I can tell. Merging this now but can revisit if anything else does come up

PR #417, opened 2023-06-05
Custom Drill Costs for Geothermal

ppinchuk:
This PR adds the ability for users to add `drill_cost_per_well` as an input to the SAM configs so that the capital cost can be adjusted accordingly. This gives much more realistic cost values for large depths (eg 4-10km).

This PR also allows users to input a nameplate value in the SAM config to override the data input.

Finally, this PR adds a fix for EGS plant temperature - this value will now be adjusted to match the temperature input from the resource data. SAM throws an error otherwise, which messes up a large portion of the EGS data points.

Labeling as Urgent b/c we would like to get one of the analysist running some supply curves with these changes this week.

### DIFF
--- a/reV/SAM/defaults.py
+++ b/reV/SAM/defaults.py
@@ -48,8 +48,11 @@ class AbstractDefaultFromConfigFile:
             config = json.load(f)
 
         for k, v in config.items():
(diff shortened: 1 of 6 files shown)


grantbuster (approved):
Looks fine, sorry for the late review!

PR #420, opened 2023-08-18
Fix out path bug

ppinchuk:
Previous version of code would add a year value to the module name everywhere in the path.
For example, a generation run for year 2012 in the directory  `/my/path/to/generation/runs` would give an output file in `/my/path/to/generation_2012/run/generation_2012.h5`. New version of code only performs this substitution in the file name.

### DIFF
--- a/reV/generation/base.py
+++ b/reV/generation/base.py
@@ -869,25 +869,26 @@ class BaseGen(ABC):
         if out_fpath is None:
             return
 
-        # ensure output file is an h5
-        if not out_fpath.endswith('.h5'):
-            out_fpath += '.h5'
+        project_dir, out_fn = os.path.split(out_fpath)
 
-        if module not in out_fpath:
+        # ensure output file is an h5
+        if not out_fn.endswith('.h5'):
+            out_fn += '.h5'
+
+        if module not in out_fn:
             extension_with_module = "_{}.h5".format(module)
-            out_fpath = out_fpath.replace(".h5", extension_with_module)
+            out_fn = out_fn.replace(".h5", extension_with_module)
 
         # ensure year is in out_fpath
-        if self.year is not None and str(self.year) not in out_fpath:
+        if self.year is not None and str(self.year) not in out_fn:
             module_with_year = "{}_{}".format(module, self.year)
-            out_fpath = out_fpath.replace(module, module_with_year)
+            out_fn = out_fn.replace(module, module_with_year)
 
         # create and use optional output dir
-        dirout = os.path.dirname(out_fpath)
-        if dirout and not os.path.exists(dirout):
-            os.makedirs(dirout)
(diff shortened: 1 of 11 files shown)


reV/generation/base.py
```diff
@@ -869,25 +869,26 @@ def _init_fpath(self, out_fpath, module):
...
             extension_with_module = "_{}.h5".format(module)
-            out_fpath = out_fpath.replace(".h5", extension_with_module)
+            out_fn = out_fn.replace(".h5", extension_with_module)
 
         # ensure year is in out_fpath
-        if self.year is not None and str(self.year) not in out_fpath:
+        if self.year is not None and str(self.year) not in out_fn:
             module_with_year = "{}_{}".format(module, self.year)
-            out_fpath = out_fpath.replace(module, module_with_year)
+            out_fn = out_fn.replace(module, module_with_year)
```
grantbuster, line 885: I could see this replace statement being problematic if "gen" or some other module name was in the project name. Maybe safer to replace the ".h5" again?

reV/generation/base.py
```diff
@@ -869,25 +869,26 @@ def _init_fpath(self, out_fpath, module):
...
 
         # create and use optional output dir
-        dirout = os.path.dirname(out_fpath)
-        if dirout and not os.path.exists(dirout):
-            os.makedirs(dirout)
+        if project_dir and not os.path.exists(project_dir):
+            os.makedirs(project_dir)
```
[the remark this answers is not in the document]
ppinchuk: Good call, definitely adding that.

reV/generation/base.py
ppinchuk: Well, I think we *intend* to replace the module name piece of the filename here. For for example, a file named `my_project_generation.h5` should become  `my_project_generation_2012.h5` . But we also want `my_project_generation_j000.h5`  to become `my_project_generation_2012_j000.h5`. 

I think the problem you are thinking of is for a file name like `my_generation_project_generation_j000.h5`, which would become `my_generation_2012_project_generation_2012_j000.h5`.  I'll try to come up with a way to handle this... Might not look pretty though

grantbuster: yes the second paragraph you sent. Just do another replacement on the .h5 same as addition of the module name:

```
        if module not in out_fn:
            extension_with_module = "_{}.h5".format(module)
            out_fn = out_fn.replace(".h5", extension_with_module)

        # ensure year is in out_fpath
        if self.year is not None and str(self.year) not in out_fn:
            extension_with_y = "_{}.h5".format(self.year)
            out_fn = out_fn.replace(".h5", extension_with_y)

ppinchuk: Gotcha. Yes, will definitely add that

grantbuster (approved):

PR #423, opened 2023-08-23
Power Curve shift compute function

ppinchuk:
Refactored power curve losses logic to extract a function (`adjust_power_curve`) that can be used to compute the power curve shift for a single site. Note that no logic changes were introduced, and therefore no new tests were added. 

I have had this kind of request (compute power curve shift for a single site) come through a few times now, and this PR adds the functionality for this kind of computation straight into reV.

An example of this process was added to the reV losses demo and example RST file.


reV/losses/power_curve.py
```diff
@@ -505,10 +502,153 @@ def target(self):
...
+        References
+        ----------
+        .. [1] Scaling done in SAM: https://tinyurl.com/2uzjawpe
+        .. [2] SAM Wind Power Reference Manual for explanations on
+           generation and air density calculations (pp. 18):
+           https://tinyurl.com/2p8fjba6
+
+        """
+
+        pressures_pascal = self._pressures * 101325.027383  # originally in atm
```
grantbuster, line 557: Worth doing a simple check to see if p is in atm or Pa? WTK is in Pa. I know docstring says pressure should be in ATM but still...

grantbuster (approved):
Minor comment on pressure units. Can ignore if you want.

reV/losses/power_curve.py
ppinchuk: I would love this actually. Is there a good way to distinguish the unit type from the value alone? Just check for values over 1e6?

grantbuster: Yeah i typically just set a threshold... Pa should be >1e4 and atm should be less than 2. These are generous but you can get some low pressures in very mountainous regions and whatnot.

grantbuster (approved):

PR #424, opened 2023-09-05
fixed bug where power curve losses was shifting power curve to the le…

grantbuster:
…ft to acheive losses

### DIFF
--- a/tests/test_losses_power_curve.py
+++ b/tests/test_losses_power_curve.py
@@ -24,7 +24,7 @@ from reV.losses.power_curve import (PowerCurve, PowerCurveLosses,
                                     PowerCurveLossesInput,
                                     TRANSFORMATIONS,
                                     HorizontalTranslation,
-                                    AbstractPowerCurveTransformation
+                                    AbstractPowerCurveTransformation,
                                     )
 from reV.losses.scheduled import ScheduledLossesMixin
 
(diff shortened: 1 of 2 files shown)


tests/test_losses_power_curve.py
```diff
@@ -341,13 +341,27 @@ def test_transformation_classes_apply(pc_transformation, real_power_curve):
...
+    new_pc_gen = copy.deepcopy(real_power_curve.generation)
+    new_pc_gen[-10:] = 0
+    new_power_curve = PowerCurve(real_power_curve.wind_speed, new_pc_gen)
+    transformation = pc_transformation(new_power_curve)
     new_power_curve = transformation.apply(strength)
     mask = new_power_curve.wind_speed >= real_power_curve.wind_speed[-10]
     assert (new_power_curve[mask] == 0).all()
 
 
+@pytest.mark.parametrize('TransClass', TRANSFORMATIONS.values())
```
ppinchuk, line 354: That is a CLEAN test. And tests any future transformations too. Love it.

ppinchuk (approved):
I tried to poke holes in this but it seems to cover all the cases I could think of so good job! Maybe real data out there will reveal something I overlooked but we can cross that bridge when we get there

grantbuster:
> I tried to poke holes in this but it seems to cover all the cases I could think of so good job! Maybe real data out there will reveal something I overlooked but we can cross that bridge when we get there

Just finished CONUS onshore+offshore runs for both Sup3rCC datasets with 0 errors or warnings, i'm encouraging! going to merge.

PR #425, opened 2023-09-15
Add warning about `pass_through_dsets`

ppinchuk:
@grantbuster Can you let me know if the warning text makes sense? If not, what would be a good way to rephrase it?

### DIFF
--- a/reV/handlers/multi_year.py
+++ b/reV/handlers/multi_year.py
@@ -64,6 +64,16 @@ class MultiYearGroup:
             multi-year files (e.g. input datasets that don't vary from
             year to year) that should be copied to the output multi-year
             file once without a year suffix or means/stdev calculation.
+
+            .. WARNING:: Unlike the `dsets` input, datasets in
+               `pass_through_dsets` **will not** be overwritten if the
+               multi-year output file exists and contains the requested
+               dataset already. This means repeated calls to multi-year
+               may leave old data byproducts that don't match the newer
+               single-year files if the multi-year output file was not
+               removed between executions. For best results, always
+               remove the multi-year output between runs.
+
             By default, ``None``.
         """
         self._name = name


grantbuster (dismissed):
It makes sense but i think it's a surprising result. Should we just be overwriting the datasets if multi year is rerun?

ppinchuk:
It's definitely a surprising result - one which caused quite a bit of confusion this past week. 

I _think_ the reason we don't overwrite pass-through datasets is to avoid repeatedly writing the same (necessarily identical) pass-through dataset from each single-year file during processing. In other words, we indiscriminately call the [_copy_dset](https://github.com/NREL/reV/blob/10e39c553799d1f3a1a2f279798eef2358d9310e/reV/handlers/multi_year.py#L301) function on every dataset of every single-year generation file, and expect it not to do any extra work on those identical pass-through datasets.

Now that I think about it, this may actually cause problems with the `dsets` inputs as well if re-running the multi-year step with an existing multi-year gen file. But perhaps the simplest solution to this problem is to ask the user to delete the file before re-running? 

Curious to hear your thoughts @grantbuster

ppinchuk:
I guess that or automatically purge any existing multi-year file at the beggining of processing. But I wonder if purging user data would open it's own can of worms...

grantbuster:
> I guess that or automatically purge any existing multi-year file at the beggining of processing. But I wonder if purging user data would open it's own can of worms...

My preference would be to purge any target output file if it already exists... imagine if jobs failed while halfway done collecting data. You would get garbage results without warning because the dataset would have initialized but not finished writing. Probably a dramatic and rare example but possible. For collect and multi-year i think this makes sense. For Gen i think we probably overwrite datasets regardless because we init first then write later?

ppinchuk:
Yes, I tend to agree. Generation just overwrite the files completely, so purging the MY file before processing makes things more consistent. I've updated the code accordingly

grantbuster:
> Yes, I tend to agree. Generation just overwrite the files completely, so purging the MY file before processing makes things more consistent. I've updated the code accordingly

Should we check collect while we're at it?

ppinchuk:
The collect logic is handled in GAPS, so we'd have to make changes there. Looks like you guys already thought about this and added a ["clobber" option](https://github.com/NREL/gaps/blob/66b8aaee1947c49865787a1f8e600115b4890e1f/gaps/collection.py#L406). However, looks like this was set to `False` by default and not exposed to the user for some reason.
I'll go ahead and expose it to the CLI, and update the default value to `True`, thereby giving the user full control while also maintaining reasonable default behavior

ppinchuk:
I guess we can add a "clobber" options with a default `True` value to MY. What do you think?

ppinchuk:
If you really are indifferent, I'll probably go ahead and add it. A default `True` option will purge files (without any extra user intervention) in all cases we can think of at the moment where files _should_ be clobbered. But as soon as an exception to this rule inevitably shows up, the power will already be in the users hands :)

grantbuster:
> If you really are indifferent, I'll probably go ahead and add it. A default `True` option will purge files (without any extra user intervention) in all cases we can think of at the moment where files _should_ be clobbered. But as soon as an exception to this rule inevitably shows up, the power will already be in the users hands :)

Yep sounds good to me; I am truly indifferent.

ppinchuk:
Sounds good. Added that in

ppinchuk:
@grantbuster any other requests or am I good to merge?

grantbuster (approved):
all good!

PR #435, opened 2023-11-21
Generation `max_workers` defaults to `None` if omitted from config file

ppinchuk:
Mostly a QOL change - it has no impact if the user explicitly specifies `max_workers` in their config file. However, if the user _omits_ this input from their config file altogether, the default value is set to `None` instead of `1`. The default value for the API is still `1`.

### DIFF
--- a/reV/generation/cli_gen.py
+++ b/reV/generation/cli_gen.py
@@ -43,6 +43,7 @@ def _preprocessor(config, job_name, log_directory, verbose,
         Updated config file.
     """
     init_cli_logging(job_name, log_directory, verbose)
+    config.get("execution_control", {}).setdefault("max_workers")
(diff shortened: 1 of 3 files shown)


grantbuster (approved):
Nice way to preserve the config default vs. python default, but where does the max_workers=None default get set? I see that `.setdefault()` call but does that assume None? Is there a corresponding GAPs PR?

ppinchuk:
Yea the `setdefault()` function will add the key to the dictionary if it's missing with a value of `None` by default. If the key _is_ in the dictionary, the function just returns the value and leaves the dictionary unchanged. 

No need for any new GAPs functionality - we simply perform a pre-processing check on the config and force a default value if needbe.

PR #436, opened 2023-12-05
Degradation

WilliamsTravis:
This adds the ability to get lifetime outputs, though with restrictions on which variables may be returned. Definitely open to suggestion about how best to handle the limited available lifetime outputs from SAM.

### DIFF
--- a/reV/generation/generation.py
+++ b/reV/generation/generation.py
@@ -12,18 +12,24 @@ import numpy as np
 import pandas as pd
 
 from reV.generation.base import BaseGen
-from reV.utilities.exceptions import (ProjectPointsValueError, InputError)
-from reV.SAM.generation import (Geothermal,
-                                PvWattsv5,
-                                PvWattsv7,
-                                PvWattsv8,
-                                PvSamv1,
-                                TcsMoltenSalt,
-                                WindPower,
-                                SolarWaterHeat,
-                                TroughPhysicalHeat,
-                                LinearDirectSteam,
-                                MhkWave)
+from reV.utilities.exceptions import (
+    ConfigError,
+    InputError,
+    ProjectPointsValueError,
+)
+from reV.SAM.generation import (
+    Geothermal,
+    PvWattsv5,
+    PvWattsv7,
+    PvWattsv8,
+    PvSamv1,
+    TcsMoltenSalt,
+    WindPower,
+    SolarWaterHeat,
+    TroughPhysicalHeat,
+    LinearDirectSteam,
+    MhkWave
+)
 from reV.utilities import ModuleName
 
 from rex.resource import Resource
@@ -51,18 +57,19 @@ class Gen(BaseGen):
     """Gen"""
 
     # Mapping of reV technology strings to SAM generation objects
-    OPTIONS = {'geothermal': Geothermal,
-               'pvwattsv5': PvWattsv5,
-               'pvwattsv7': PvWattsv7,
-               'pvwattsv8': PvWattsv8,
-               'pvsamv1': PvSamv1,
-               'tcsmoltensalt': TcsMoltenSalt,
-               'solarwaterheat': SolarWaterHeat,
-               'troughphysicalheat': TroughPhysicalHeat,
-               'lineardirectsteam': LinearDirectSteam,
-               'windpower': WindPower,
-               'mhkwave': MhkWave
-               }
+    OPTIONS = {
+        'geothermal': Geothermal,
+        'pvwattsv5': PvWattsv5,
+        'pvwattsv7': PvWattsv7,
+        'pvwattsv8': PvWattsv8,
+        'pvsamv1': PvSamv1,
+        'tcsmoltensalt': TcsMoltenSalt,
+        'solarwaterheat': SolarWaterHeat,
+        'troughphysicalheat': TroughPhysicalHeat,
+        'lineardirectsteam': LinearDirectSteam,
+        'windpower': WindPower,
+        'mhkwave': MhkWave
+    }
     """reV technology options."""
 
     # Mapping of reV generation outputs to scale factors and units.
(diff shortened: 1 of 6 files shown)


ppinchuk:
Gunna have to get those tests fixed before we do much more with this I think

WilliamsTravis:
Hi, yes definitely, got pulled away by a more pressing issue with the runs i need this for, will fix asap.

reV/generation/generation.py
```diff
@@ -12,18 +12,24 @@
...
-                                PvWattsv7,
-                                PvWattsv8,
-                                PvSamv1,
-                                TcsMoltenSalt,
-                                WindPower,
-                                SolarWaterHeat,
-                                TroughPhysicalHeat,
-                                LinearDirectSteam,
-                                MhkWave)
+from reV.utilities.exceptions import (
```
ppinchuk, line 15: Can we retain original formatting here and anywhere else in the PR where a pure formatting change was made?

reV/generation/generation.py
```diff
@@ -482,11 +489,85 @@ def time_index(self):
...
+        ----------
+        ti : pandas.DatetimeIndex
+            Time-series datetime index with leap days.
+
+        Returns
+        -------
+        ti : pandas.DatetimeIndex
+            Time-series datetime index.
+        """
+        life_var = "system_use_lifetime_output"
```
ppinchuk, line 513: Can you please document this key and it's subtleties (i.e. every SAM config has to have this option enabled, and every SAM config has to have the same `analysis_period`, etc.) after the `clearsky` documentation here: https://github.com/NREL/reV/blob/0bf8745d5725e83d49a6a67570996a11d4a0f1aa/reV/SAM/generation.py#L817-L820

reV/generation/generation.py
```diff
@@ -482,11 +489,85 @@ def time_index(self):
...
+        lifetime_periods = []
+        for sam_meta in self.sam_metas.values():
+            if life_var in sam_meta and sam_meta[life_var] == 1:
+                lifetime_period = sam_meta["analysis_period"]
+                lifetime_periods.append(lifetime_period)
+            else:
+                lifetime_periods.append(1)
+
+        if any(ltp > 1 for ltp in lifetime_periods):
```
ppinchuk, line 522: Can you please invert this check to be
```suggestion
        if not any(ltp > 1 for ltp in lifetime_periods):
            return ti
```
Then all the code below it can come down one indent, which is preferable

reV/generation/generation.py
```diff
@@ -482,11 +489,85 @@ def time_index(self):
...
+            array_vars = [
+                var for var, attrs in GEN_ATTRS.items()
+                if attrs['type'] == 'array'
+            ]
+            valid_vars = ['gen_profile', 'cf_profile', 'cf_profile_ac']
+            invalid_vars = set(array_vars) - set(valid_vars)
+            invalid_requests = [var for var in self.output_request
+                                if var in invalid_vars]
+
+            if n_unique_periods != 1:
```
ppinchuk, line 534: Minor, but can you please move this check block to immediately after `n_unique_periods` is defined?

reV/generation/generation.py
```diff
@@ -482,11 +489,85 @@ def time_index(self):
...
+                # Only one time index, so lifetime outputs all need to match
+                msg = ('reV cannot handle multiple analysis_periods when '
+                       'modeling with `system_use_lifetime_output` set '
+                       'to 1. Found {} different analysis_periods in the SAM '
+                       'configs'.format(n_unique_periods))
+                logger.error(msg)
+                raise ConfigError(msg)
+
+            elif invalid_requests:
```
ppinchuk, line 543: The `elif` is not necessary here, since code execution terminates when the exception in the above block is raised. Can you please change it to a simple `if` and move it to be just under where `invalid_requests` is defined (should happen automatically once you move the block right above)

reV/generation/generation.py
```diff
@@ -482,11 +489,85 @@ def time_index(self):
...
+                    'reV can only handle the following output arrays '
+                    'when modeling with `system_use_lifetime_output` set '
+                    'to 1: {}. Try running without {}.'.format(
+                        ', '.join(valid_vars), ', '.join(invalid_requests)
+                    )
+                )
+                logger.error(msg)
+                raise ConfigError(msg)
+
+            else:
```
ppinchuk, line 555: The `else` is also not necessary, for the same reason as above. Please remove it and un-indent the block below

ppinchuk (changes requested):
Nice job! A few minor requests, with special emphasis on the formatting thorughout.
Huge kudos for doing all the validation checks AND including tests. You're basically a reV developer now :)

reV/generation/generation.py
ppinchuk: Ah, I see this is a SAM key, not just a reV-specific key. In that case, can you create a section in the style that we have for reV Geothermal ("notable SAM keys"), and put the documentation there instead? https://github.com/NREL/reV/blob/0bf8745d5725e83d49a6a67570996a11d4a0f1aa/reV/SAM/generation.py#L1324-L1331

WilliamsTravis: Okay, wrote a description of the lifetime specific keys needed, check it out

reV/generation/generation.py
WilliamsTravis: Done

reV/generation/generation.py
WilliamsTravis: Done

reV/generation/generation.py
WilliamsTravis: Done

reV/generation/generation.py
WilliamsTravis: Done

WilliamsTravis:
Now it is failing on docstring lines that are over the line length limit, though these were there before and it had passed. Not sure how you'd like to handle this.

reV/generation/generation.py
ppinchuk: Looks great, thanks so much!!

ppinchuk:
Ooops! Those are totally my bad. Not much to do about it though, since we need that length for the link to work (I'm pretty sure?). Anyways, I disabled the check for those docstrings

ppinchuk (approved):
Nice job!

PR #439, opened 2024-01-10
Gb/bc

grantbuster:
Primarily motivated by Sup3rCC, we need to be able to use more complex bias correction methods so this PR opens up some flexibility in using any functions we implement in rex.


ppinchuk:
Looks like we should handle the rex PR first, publish it to PyPi, then bump the req here for the tests?

ppinchuk (approved):
Yea, nice job! 
Tried to poke holes in it but couldn't find any.
Thanks for the special attention to the docstrings :)

PR #442, opened 2024-02-13
Gb/bespoke bugs

grantbuster:
Basically we never tested bespoke from a prior run with LCOE calculations and there was a bug with this combination. Simple fix though.

### DIFF
--- a/reV/bespoke/bespoke.py
+++ b/reV/bespoke/bespoke.py
@@ -803,6 +803,11 @@ class BespokeSinglePlant:
                                          'pressure': pres,
                                          'windspeed': ws,
                                          'winddirection': wd}, index=ti)
+
+            if 'time_index_step' in self.original_sam_sys_inputs:
+                ti_step = self.original_sam_sys_inputs['time_index_step']
+                self._res_df = self._res_df.iloc[::ti_step]
+
         return self._res_df
 
     @property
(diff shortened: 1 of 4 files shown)


reV/bespoke/bespoke.py
```diff
@@ -969,27 +975,36 @@ def get_lcoe_kwargs(self):
...
-            'variable_operating_cost':
-                self.plant_optimizer.variable_operating_cost}
+        kwargs_list = ['fixed_charge_rate', 'system_capacity', 'capital_cost',
+                       'fixed_operating_cost', 'variable_operating_cost']
+        lcoe_kwargs = {}
+
+        for kwarg in kwargs_list:
+            if kwarg in self.outputs:
+                lcoe_kwargs[kwarg] = self.outputs[kwarg]
+            elif getattr(self.plant_optimizer, kwarg, None) is not None:
```
ppinchuk, line 989: Look at this wonderful opportunity for our favorite marine mammal!! 

image

ppinchuk (approved):
Yea, nice and clean.

I know the time-step fix is trying to go out the door ASAP so I understand no tests, but have you tried running it at all to see if it works? I knew it would be a simple fix but it's kind of crazy that it can be handled with just three little lines!

ppinchuk (approved):

PR #447, opened 2024-03-05
Report scaled capital cost in output supply curve

ppinchuk:
When specifying an EOS curve, we were reporting the EOS multiplier and the scaled LCOE but not the scaled capital cost. This PR now also outputs the scaled capital cost value in the SC, which is needed for downstream models like ReEDS.

Also report cost values and annual energy for each supply curve based on its capacity.


ppinchuk:
@grantbuster this PR is based on what we discussed this morning. Let me know if you see any red flags or if I am missing anything here

grantbuster (approved):
This looks great, thanks for catching this and making the code more readable for future us!

PR #449, opened 2024-04-26
Exclusion extent functionality

ppinchuk:
Add the ability to specify an extent that a particular exclusion can be applied to.

The extent can be specified in the same way that any other exclusion layer is, but it must come out to be a boolean layer.

Fully recursive excl dict was explored as part of this update, but I could not figure out how to make it work properly without a giant refactor, so am tabling that idea for now. Happy to explore that separately.


ppinchuk:
@grantbuster No need to do a full review if you are busy (Mike and I both already tested this on a full CONUS sample), but I wanted to give you the chance to comment if you wanted to before I merge this

PR #457, opened 2024-06-24
Add resource length warning for easier debugging

ppinchuk:
@grantbuster Reid recently ran into a tough-to-debug issue where SAM wasn't setting the output variables correctly, and it came down to the fact that he was using `time_index_step=2` on hourly resource data.

Since this was a tough problem to track down, I figured I would add a few warning messages if this case is detected. Any objections? Any reason why someone would want to run less than a full 8760 time series through SAM? Is that even allowed (seems like not based on this experience but I haven't fully tested that)?

### DIFF
--- a/reV/SAM/generation.py
+++ b/reV/SAM/generation.py
@@ -234,11 +234,10 @@ class AbstractSamGeneration(RevPySam, ScheduledLossesMixin, ABC):
             raise InputError(msg)
 
         if len(resource) < 8760:
-            msg = (f"Detected resource time series of length "
-                   f"{len(resource)}, which is less than 8760. This may "
-                   f"yeild unexpected results or fail altogether. If this "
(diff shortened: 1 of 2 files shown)


grantbuster (dismissed):
Good addition. Maybe note "try setting 'time_index_step: 1' in your SAM config or _double check the resource input you're using_"

In the future i think there will be value in running monthly data to generate profiles for event analysis. SAM used to not be able to do that but i'm pretty sure now it can just the annual values will not be set (annual CF).

ppinchuk:
Ah good to know. Interested to see if reV can handle the monthly case out of the box or not. We should probably set up a test for that when it becomes relevant - might be a good time to re-evaluate this message too :) 

(Happy to take charge on that whenever it comes around)

PR #461, opened 2024-08-13
Allow bespoke to read list for resource input

ppinchuk:
Allow bespoke class to ingest list as resource file, so that you can use

```JSON
"res_fpath": [
    "/datasets/WIND/conus/v1.0.0/wtk_conus_*.h5",
    "/datasets/WIND/HRRR/hrrr_conus_*.h5",
]
```

in your bespoke config. Note that vanilla reV instead requires you to list the files out individually. This is because vanilla generation allows brackets `{}` in the file path that would get filled in by the `analysis_years` input, so it would be ambiguous which years to use for which file path if the input was:

```JSON
"resource_file": [
    "/datasets/WIND/conus/v1.0.0/wtk_conus_{}.h5",
    "/datasets/WIND/HRRR/hrrr_conus_{}.h5",
]
```
(i.e. does 2007 below with WTK or HRRR? What about 2015?, etc).

This PR includes tests for list input, but this use pattern was also tested in production for FY24 SS with WTK+BCHRRR runs.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,2 @@
+# GitHub syntax highlighting
+pixi.lock linguist-language=YAML linguist-generated=true
--- a/.gitignore
+++ b/.gitignore
@@ -116,3 +116,8 @@ tags
 # Scratch and temp work
 .tmp/
 .scratch/
+
+
+# pixi environments
+.pixi
+*.egg-info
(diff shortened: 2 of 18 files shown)


ppinchuk:
@grantbuster This PR moves to using `MultiYearWindResource` exclusively in bespoke. I haven't run across any issues in doing so in the tests or production. Any edge case you could think of where we would definitely want to use the basic `WindResource` file handler?

PR #477, opened 2024-10-22
Fix regional cost bug

ppinchuk:
Previously, regional multipliers would be applied twice - once during the generation run and once during the econ run. We need the econ data to be formalized at generation run time though, for output collection purposes. So, in this PR, I added a flag that let's reV know if econ values have already been adjusted. If so, no other adjustments are made.

A test for regional multipliers did exist, but it was only on the econ module. This bug happens when generation and econ are both run so this bug slipped through the cracks. I added a new test that checks the multipliers for a combined generation + econ run


ppinchuk:
reVX test fails because of pandas 2.0. Moving forward with this for now

PR #478, opened 2024-10-23
Fix MY handler regression

ppinchuk:
If user defines "dsets" in MY collection config as a subset of all possible dsets, then some required LCOE dsets may end up missing from the output (i.e. not added to `pass_through_dsets`). In this PR, we pull the full set of available datasets regardless of the user input. Includes test.


ppinchuk:
reVX test fails because of pandas 2.0. Moving forward with this for now

PR #502, opened 2025-01-10
Gb/revs3

grantbuster:
Enables reV to run directly from filepaths on s3 without HSDS setup. Will be less performant but will require zero setup. Also updated examples and included a test.

### DIFF
--- a/.github/workflows/s3_tests.yml
+++ b/.github/workflows/s3_tests.yml
@@ -1,0 +1,31 @@
+name: s3 fsspec tests
+
+on: pull_request
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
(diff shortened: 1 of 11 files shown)


ppinchuk (dismissed):
Thanks for this!

.github/workflows/s3_tests.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Install dependencies
+        shell: bash
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install pytest
+          python -m pip install .[s3]
+      - name: Run pytest and Generate coverage report
+        shell: bash
+        run: |
+          python -m pytest -v --disable-warnings
```
ppinchuk, line 34: Is this now missing `s3_tests.py`?

grantbuster: oh dammit. good catch!

PR #503, opened 2025-01-16
Supply Curve Aggregation by Zone within GIDs

mjgleason:
First cut at implementing supply curve aggregation for zones within GIDs. This functionality is enable by passing values to the new `zones_dset` argument. 

I added new tests to `tests/test_supply_curve_sc_aggregation.py`, and the existing tests are all passing as well. I was unable to directly test that this functionality works with `cap_cost_scale` because the test gen/econ datasets do not have the correctly named economic datasets, but I did test that each zone + sc site is output to a separate sc_gid with the correct areas and capacities. Since the capital cost scale functionality is already tested separately, and runs downstream of the area and capacity calculations for each site in `GenerationSupplyCurvePoint.summarize()`, I don't this its okay not to test it. 

This PR also includes some incidental changes to `.pylintrc` and `.pre-commit-config.yaml` to get them working again (they were pretty outdated).


reV/supply_curve/aggregation.py
```diff
@@ -438,6 +438,53 @@ def _get_gid_inclusion_mask(
...
+                    )
+                    logger.error(msg)
+                    raise FileInputError(msg)
+                zones = fh[zones_dset, row_slice, col_slice]
+        else:
+            shape = (
+                row_slice.stop - row_slice.start,
+                col_slice.stop - col_slice.start
+            )
+            zones = np.ones(shape, dtype="uint32")
```
ppinchuk, line 484: Why not "uint8"?

reV/supply_curve/points.py
```diff
@@ -454,6 +462,24 @@ def n_gids(self):
...
+        Get the 2D zone mask, where 1 is included and 0 is excluded.
+
+        Returns
+        -------
+        np.ndarray
+        """
+
+        if self._zone_mask is None:
+            return None
+        else:
```
ppinchuk, line 477: No need for the else here I think

reV/supply_curve/points.py
```diff
@@ -454,6 +462,24 @@ def n_gids(self):
...
+            out_of_extent = self._gids.reshape(self._zone_mask.shape) == -1
```
ppinchuk, line 478: Is this redundant? `zone_mask` is only used to zero out values in the `_inclusion_mask`, which alreasy zeros out out-of-bounds gids. Maybe I am missing something?

reV/supply_curve/points.py
```diff
@@ -548,6 +578,31 @@ def _check_excl(self):
...
+            )
+            assert len(self._zone_mask.shape) == 2, msg
+            assert self._zone_mask.shape[0] <= self._resolution, msg
+            assert self._zone_mask.shape[1] <= self._resolution, msg
+            assert self._zone_mask.size == len(self._gids), msg
+
+            if not np.isin(self._zone_mask, [0, 1]).all():
+                msg = (
+                    "zone_mask includes unexpected values. All values must be "
+                    "in the domain: [0, 1]"
```
ppinchuk, line 601: This is a bit misleading since zone mask is actually a boolean array that I assume is implicitly cast to ints during the `np.isin` operation (and probably during the `self.zone_mask == 0` line above)

ppinchuk (commented):
Leaving comments here now so that I don't forget them over the weekend

reV/supply_curve/points.py
ppinchuk: Yup! I think returning `self._zone_mask` is perfect here

reV/supply_curve/points.py
ppinchuk: Looks good to me!

reV/supply_curve/sc_aggregation.py
```diff
@@ -1093,45 +1112,56 @@ def run_serial(
...
+                                zone_mask=zone_mask,
+                            )
 
-                        summary.append(pointsum)
-                        logger.debug(
-                            "Serial aggregation completed gid {}: "
-                            "{} out of {} points complete".format(
-                                gid, n_finished, len(gids)
+                        except EmptySupplyCurvePointError:
+                            logger.debug("SC point {} is empty".format(gid))
```
ppinchuk, line 1151: Leaving this here now so I don't forget later. Let's add the zone ID to this logger message, otherwise the logs look something like this: 
image

ppinchuk (commented):
Quick comment

reV/supply_curve/sc_aggregation.py
ppinchuk: Added this since I was already working in the code

reV/supply_curve/sc_aggregation.py
```diff
@@ -1093,45 +1112,56 @@ def run_serial(
...
+                            pointsum['zone_id'] = zone_id
+
+                            summary.append(pointsum)
+                            logger.debug(
+                                "Serial aggregation completed gid {}, "
+                                "resource class {}, zone ID {}: "
+                                "{} out of {} points complete".format(
+                                    gid, ri, zone_id, n_finished,
+                                    len(gids)
+                                )
```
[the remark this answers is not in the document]
ppinchuk: Added this since I was already working in the code

reV/supply_curve/points.py
```diff
@@ -282,6 +288,18 @@ def __init__(
...
+        if zone_mask is not None:
+            msg = (
+                "Bad zone mask input shape of {} with stated "
+                "resolution of {}".format(zone_mask.shape, resolution)
+            )
+            assert len(zone_mask.shape) == 2, msg
+            assert zone_mask.shape[0] <= resolution, msg
+            assert zone_mask.shape[1] <= resolution, msg
+            assert zone_mask.size == len(self._gids), msg
+            self._zone_mask = zone_mask.copy()
```
ppinchuk, line 301: This copy seems to be intentionally placed, but I can't seem to figure out why. Do you ever modify the values within? Did you run into some sort of issues without the copy? or is this just trying to be future-proof?

tests/test_supply_curve_sc_aggregation.py
```diff
@@ -711,6 +712,203 @@ def test_format_res_fpath_with_year_pattern():
...
+            attrs = dict(f['ri_smod'].attrs)
+            profile = json.loads(attrs["profile"])
+            profile["dtype"] = "uint32"
+            profile["nodata"] = 0
+            attrs["profile"] = json.dumps(profile)
+            data = np.zeros(shape, dtype=np.uint32)
+            if zone_config == "one_full":
+                # each entire cell is one zone
+                for gid, gid_slice in slice_lookup.items():
+                    data[gid_slice] = gid + 10
```
ppinchuk, line 746: Would it be worth having a separate test with multiple zones per SC point? Just to make sure we get repeated `sc_point_gid` outputs with different zone ID's?  Or is that too cumbersome to set up?

ppinchuk (commented):
This is awesome, thank you! Couple of quick questions and then I think we can get this merged!

reV/supply_curve/points.py
ppinchuk: Ok, the copy here is probably not necessary, but let's not have it hold up the merge. We can drop it later if it becomes a problem

tests/test_supply_curve_sc_aggregation.py
ppinchuk: Oooops! I guess I missed the other branch of the if statement somehow. 
Thanks for being so thorough!

ppinchuk (approved):

PR #504, opened 2025-01-23
Decreasing memory footprint of Tech Mapping

mjgleason:
This PR decreases the memory footprint of Tech Mapping to enable creation of tech maps for exclusions at higher spatial resolution (e.g., 30m instead of 90m). It does so primarily by writing the results in slices to the h5 dataset, rather than building it up as an array in memory and then writing the full array to disk. There are a few other memory-saving changes as well.

The PR also introduces a new CLI command `tech-mapping` which allows running the creation of tech maps in isolation from running supply curve aggregation.

This PR is technically breaking since the `TechMapping.run()` method used to return the tech map array. This functionality did not appear to be used anywhere (in the codebase or the tests), so I removed it since it more or less defeats the purpose of trying to optimize memory.

**TODOS:**
- [x] revert to `sc_resolution` and `dset` parameters (instead of `resolution` and `tm_dset`) to avoid unnecessary breaking changes to the API
- [x] expose `batch_size` as parameter up the call chain

### DIFF
--- a/reV/supply_curve/tech_mapping.py
+++ b/reV/supply_curve/tech_mapping.py
@@ -31,7 +31,7 @@ class TechMapping:
     """Framework to create map between tech layer (exclusions), res, and gen"""
 
     def __init__(
-        self, excl_fpath, res_fpath, sc_resolution=64, dist_margin=1.05
+        self, excl_fpath, res_fpath, sc_resolution=1200, dist_margin=1.05
     ):
         """
         Parameters
@@ -42,11 +42,16 @@ class TechMapping:
         res_fpath : str
             Filepath to .h5 resource file that we're mapping to.
         sc_resolution : int | None, optional
-            Supply Curve resolution. This value defines how many pixels
-            are in a single side of a supply curve cell. For example,
-            a value of ``64`` would generate a supply curve where the
-            side of each supply curve cell is ``64x64`` exclusion
-            pixels. By default, ``64``.
+            Defines how many exclusion pixels are mapped at a time. Units
+            indicate the length of one dimension, in pixels, of each square
+            chunk to be mapped. By default, this value is 1200, which will
+            map the exclusion pixels in 1200x1200 pixel chunks.
+
+            .. Note:: This parameter does not affect the exclusion to resource
+            (tech) mapping, which deviates from how the effect of the
+            ``sc_resolution`` parameter works in other functionality within
+            ``reV``.
+
         dist_margin : float, optional
             Extra margin to multiply times the computed distance between
             neighboring resource points, by default 1.05
@@ -184,16 +189,17 @@ class TechMapping:
         return row_slice, col_slice
 
     @classmethod
-    def _get_excl_coords(cls, excl_fpath, gids, sc_row_indices, sc_col_indices,
+    def _get_excl_coords(cls, excl_fpath, gid, sc_row_indices, sc_col_indices,
                          excl_row_slices, excl_col_slices,
                          coord_labels=(LATITUDE, LONGITUDE)):
         """
-        Extract the exclusion coordinates for the desired gids for TechMapping.
+        Extract the exclusion coordinates for the desired supply curve point
+        gid for TechMapping.
 
         Parameters
         ----------
-        gids : np.ndarray
-            Supply curve gids with tech exclusion points to map to the
+        gid : int
+            Supply curve gid with tech exclusion points to map to the
             resource meta points.
         excl_fpath : str
             Filepath to exclusions h5 file, must contain latitude and longitude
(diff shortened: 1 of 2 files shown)


reV/supply_curve/tech_mapping.py
```diff
@@ -393,105 +351,134 @@ def _check_fout(self):
...
+                    futures[
+                        exe.submit(
+                            self.map_resource_gids,
+                            gid_set,
+                            excl_coords,
+                            self._tree,
+                            self.distance_threshold,
+                        )
+                    ] = i
+                n_submitted += batch_size
```
ppinchuk, line 418: If we go with the `for` loop, this can be scrapped
```suggestion
```

reV/supply_curve/tech_mapping.py
```diff
@@ -393,105 +351,134 @@ def _check_fout(self):
...
+            the exclusions will be mapped. Can refer to a single file (e.g.,
+            "/path/to/nsrdb_2024.h5" or a wild-card e.g.,
+            "/path/to/nsrdb_{}.h5")
+        dset : str
+            Dataset name in the `excl_fpath` file to which the the
+            techmap (exclusions-to-resource mapping data) will be saved.
+
+            .. Important:: If this dataset already exists in the h5 file,
+              it will be overwritten.
         sc_resolution : int | None, optional
```
ppinchuk, line 476: I think for proper sphinx formatting, we'll need a newline here
```suggestion
            .. Important:: If this dataset already exists in the h5 file,
               it will be overwritten.

        sc_resolution : int | None, optional
```

tests/test_supply_curve_tech_mapping.py
```diff
@@ -23,13 +28,78 @@
...
 
     with ExclusionLayers(EXCL) as ex:
         ind_truth = ex[TM_DSET]
 
+    with ExclusionLayers(excl_fpath) as out:
+        assert dset in out, "Techmap dataset was not written to H5"
+        ind = out[dset]
+
+    msg = 'Tech mapping failed for {} vs. baseline results.'
+    assert np.allclose(ind, ind_truth), msg.format('index mappings')
```
ppinchuk, line 53: Am I crazy or can we skip the message formatting here?

```suggestion
    msg = 'Tech mapping failed for index mappings vs. baseline results.'
    assert np.allclose(ind, ind_truth), msg
```

tests/test_supply_curve_tech_mapping.py
```diff
@@ -23,13 +28,78 @@
...
+
+    with ExclusionLayers(EXCL) as ex:
+        ind_truth = ex[TM_DSET]
+
+    with ExclusionLayers(excl_fpath) as out:
+        assert dset in out, "Techmap dataset was not written to H5"
+        ind = out[dset]
+
     msg = 'Tech mapping failed for {} vs. baseline results.'
     assert np.allclose(ind, ind_truth), msg.format('index mappings')
```
ppinchuk, line 104:
```suggestion
    msg = 'Tech mapping failed for index mappings vs. baseline results.'
    assert np.allclose(ind, ind_truth), msg
```

tests/test_supply_curve_tech_mapping.py
```diff
@@ -23,13 +28,78 @@
...
```
ppinchuk, line 105: Awesome tests, thanks for adding these!!

ppinchuk (changes requested):
Thanks a ton for this! Great to have more flexibility with creating tech maps. 

My main question is around the `batch_size` parameter. It seems like this is the main wayu for users to limit the meory use of this procedure. If so, we should mention that in the documentation. 

Everything else is merely a suggestion, feel free to take it or leave it

ppinchuk:
@mjgleason Thanks for the clarification. I think I understand. But am I right that `batch_size` is the way users are intended to "control" how the memory is limited? Or was your intention not to let user have control over that at all?

I guess my thought is that if I have a machine with 200 TB of RAM, can I bypass this optimization by setting batch size really high?

ppinchuk:
@mjgleason Thanks for helping me work though this. If i understand the code correctly, `batch_size` does limit the number of jobs that are submitted to the pool at a time, like you said. I think we have two possible scenarios here: 

- `max_workers` < `batch_size`: In this case, `max_workers` would be the biggest lever the user can pull to control memory, since that input decides how many resource data array slices are being read into memory at the same time. In other words, if `batch_size=100` and `max_workers=10`, then only 10 processes are loading resource data into memory at a time, therefore limiting the total memory. This isn't quite the full story though, since I believe the pickled job submissions actually live in memory until they are picked up by a worker. Typically this is not a problem, but I think the cKDtree is a non-negligible amount of memory (~ 1GB IIRC). So if we set `batch_size=300` and `max_workers=10`, you could still conceivably hit OOM on a 250GB node.
- `max_workers` >= `batch_size`: I think this is the scenario I was originally envisioning, in which case I believe `batch_size` completely overrides `max_workers` and becomes the true way to manage how much memory the process is using. 

Am I misunderstanding anything? If my thinking is correct here, could we get away with removing the `batch_size` parameter and always using `max_workers` as the batch size?

ppinchuk (approved):
Looks great, thanks for spearheading this!!

PR #513, opened 2025-02-13
Spatial curtailment

ppinchuk:
Refactor curtailment to allow spatially-explicit curtailment options via project points input. The new curtailment functions very similar to the SAM config input now. 

Unfortunately this minorly breaks the existing curtailment specification method. Namely, users could previously input the curtailment config directly like so:
```
"curtailment":  {
    "dawn_dusk": "nautical",
    "months": [4, 5, 6, 7],
    "probability": 1,
    "wind_speed": 10.0,
}
```

In the updated version, users will have to assign the configuration a name, like so:
```
"curtailment":  {
    "main_curtailment": {
        "dawn_dusk": "nautical",
        "months": [4, 5, 6, 7],
        "probability": 1,
        "wind_speed": 10.0,
    }
}
```

The latter allows specifying multiple curtailment configs (which is required to perform spatially-explicit curtailment:
```
"curtailment":  {
    "main_curtailment": {
        "dawn_dusk": "nautical",
        "months": [4, 5, 6, 7],
        "probability": 1,
        "wind_speed": 10.0,
    },
    "other_curtailment": "/path/to/other/curtailment.json"
}
```

### DIFF
--- a/reV/SAM/SAM.py
+++ b/reV/SAM/SAM.py
@@ -190,7 +190,10 @@ class SamResourceRetriever:
             kwargs["icing"] = project_points.sam_config_obj.icing
             if (
                 project_points.curtailment is not None
-                and project_points.curtailment.precipitation
+                and any(
+                    config.precipitation
+                    for config in project_points.curtailment.values()
+                )
             ):
                 # make precip rate available for curtailment analysis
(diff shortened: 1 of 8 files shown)


ppinchuk:
This PR also addresses #515

PR #524, opened 2025-03-21
Move to `pyproject.toml`

ppinchuk:
Recently, a warning message started being thrown when doing `pip install -e .` with `setup.py`. In this PR, we move to a `pyproject.toml` file to avoid this warning and conform with [PEP-621](https://peps.python.org/pep-0621/).

Also drop support for 3.8 in this PR since it doesn't like the new style of specifying the license in `pyproject.toml`.


ppinchuk:
@grantbuster Not super critical or anything, but wanted to give you a chance to comment if you wanted to

Issue #505, opened 2025-02-06
New Captial Cost Title Should Say Capex not OCC


WilliamsTravis:
**Bug Description**
This isn't exactly a software bug, but the new supply curve fields have one little problem with the capital cost names (e.g., "cost_base_occ_usd_per_ac_mw", "cost_base_occ_usd_per_ac_mw"). The SAM config takes in a generic [capital cost parameter](https://nrel-pysam.readthedocs.io/en/main/modules/Lcoefcr.html#PySAM.Lcoefcr.Lcoefcr.SimpleLCOE.capital_cost) which may or may not represent overnight capital costs. "Overnight capital costs" exclude construction period financing while "capital costs" include this factor. While our main runs for the Standard Scenarios use overnight capital costs, this is a decision that is specific to the ReEDS modeling pipeline and is not always the case. With the current name, users will assume that the capital cost fields exclude construction period financing when this may not be the case.

**Potential Solution**
Align with SAM and use a generic capital cost tag to represent this field in the supply curve outputs. However, if we used "capital_cost" users wouldn't be able to split on "_" if needed like they can with the other cost fields, so perhaps just "capex" ("cost_base_capex_usd_per_ac_mw")

**Additional context**
This could cause a bit confusion if too many reV outputs have already been integrated in downstream models, so we need to be aware of that. However, I think its worth it to avoid continued confusion moving forward. 

**Charge code**
GDOP.12495.31.01.01

PR #541, opened 2025-05-04
Rename SC output column

ppinchuk:
As described in #505, having `occ` in the column name is misleading. In this PR, we update the column name to just say `cc`


ppinchuk:
Will fix linter error in next PR

PR #544, opened 2025-05-15
Gb/bug site height

grantbuster:
Surprising bug, easy solution?


ppinchuk:
Very surprising! Looks like that's already something that was addressed for the geothermal depth variable, which is very similar to wind hub heights. I guess we just didn't have tests for this so it was never caught

PR #567, opened 2025-08-13
Config files in output file attributes

ppinchuk:
reV now stores the config file as a file attribute if/when reV is run from the command line. This should help with traceability and reproducibility of results.


WilliamsTravis:
Excellent! I've wanted this for a while.

